PR #96, opened 2012-08-20
Tiled Object on Isometric maps bug fix

andyveliz:
The position of the objects in isometric maps was corrected, now the objects appear in the same position as they are in Tiled.

"Tile Objects" are now being added to the EntityPool.

### DIFF
--- a/src/entity/entity.js
+++ b/src/entity/entity.js
@@ -155,20 +155,7 @@
 
 		obj.newInstanceOf = function(prop) {
 			var name = prop.name ? prop.name.toLowerCase() : undefined;
-<<<<<<< HEAD
-			if (!name || !entityClass[name]) {
-				if(!prop.image){
-					console.error("cannot instance entity of type '" + name
-							+ "': Class not found!");
-					return null;
-				} else 
-					return new me.SpriteObject(prop.x, prop.y, prop.image);
-					
-			}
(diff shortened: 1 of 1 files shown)


src/entity/entity.js
```diff
@@ -155,15 +155,16 @@
...
+							+ "': Class not found!");
+					return null;
+				} else 
+					return new me.SpriteObject(prop.x, prop.y, prop.image);
+					
 			}
+				
 			// i should pass the entity ownProperty instead of the object itself
 			return new entityClass[name](prop.x, prop.y, prop);
 		};
```
parasyte, line 175: How about cleaning this up a bit?

``` JavaScript
obj.newInstanceOf = function(prop) {
    var name = prop.name ? prop.name.toLowerCase() : undefined;
    if (name && entityClass[name]) {
        // FIXME: I should pass the entity ownProperty instead of the object itself
        return new entityClass[name](prop.x, prop.y, prop);
    }

    // Tile objects can be created with a GID property;
    // The TMX parser will use it to create the image property.
    if (prop.image) {
        return new me.SpriteObject(prop.x, prop.y, prop.image);
    }

    console.error("Cannot instantiate entity of type '" + name
        + "': Class not found!");
    return null;
};
```

PR #305, opened 2013-11-12
added null check on localstorage

### LINKED ISSUE #304: bug
"obj.localStorage = (typeof window.localStorage !== 'undefined')"  is wrong in my phone

my window.localStorage is null

agmcleod:
Should fix #304

### DIFF
--- a/src/system/device.js
+++ b/src/system/device.js
@@ -54,7 +54,7 @@
 			}
 
 			try {
-				obj.localStorage = (typeof window.localStorage !== 'undefined');
+				obj.localStorage = (window.localStorage !== null && typeof window.localStorage !== 'undefined');
 			} catch (e) {
 				// the above generates an exception when cookies are blocked
 				obj.localStorage = false;


src/system/device.js
```diff
@@ -54,7 +54,7 @@
 			}
 
 			try {
-				obj.localStorage = (typeof window.localStorage !== 'undefined');
+				obj.localStorage = (window.localStorage !== null && typeof window.localStorage !== 'undefined');
```
parasyte, line 57: I'm guessing it would be better to accept any falsy value as "local storage not available"...

agmcleod, line 57: True, probably wouldn't hurt to add !window.localStorage as well. that's about my extent knowledge of falsey in JS anyways :)

parasyte, line 57:
``` javascript
obj.localStorage = window.localStorage ? true : false;
```

That should do it. Covering even weird cases like `0`, `false`, `NaN`, `""`, etc.

Also, a shortcut version could be:

``` javascript
obj.localStorage = !!window.localStorage;
```

Issue #327, opened 2013-12-11
remove the obsolete `me.game.remove'


obiot:
I'm turning this around in my head and cannot see the smart solution :

Right now, we have :
-  `me.game.remove()` function that allows to remove a child from its corresponding parent container. 
  - the function will by default delay the removal at the end of the current game loop, although this can also be forced through the `force` parameter. 
  - this function will then call its parent container `removeChild` function to actually destroy the object

Then we have :
- the `me.ObjectContainer.removeChild` function that can be used to remove a child from the container (directly), and that takes an additional `keepAlive` parameter if we need to keep a living reference of this object somewhere (else it got put back in the entity pool) 

The idea being that I would like to clean as much as possible the `me,game` namespace from any object related function (like I did with me.game.add, etc.. etc..), and with the two above function being somehow interdependent, finding the right for this one is not easy :)

PR #343, opened 2013-12-30
stripped out game.remove/removAll

agmcleod:
#327

Removed the me.game.remove() method. Updated documentation to use `me.game.world.removeChild`. 

@obiot wanted you to check it out first, given we should be careful on the remove function since the `me.game.remove` has the force option. Through my experience in using melon, the force is nothing really but trouble.

Testing against the different examples and some of my own it seems to be fine. The main thing is users will need to update their code to use removeChild.


tests/spec/droptarget.js
```diff
@@ -59,7 +59,6 @@
                 // create a new droptarget entity instance
                 droptarget = new Droptarget(100, 100, {width: 200, height: 200});
                 // add the test droptarget to the game
-                me.game.world.addChild(droptarget, 1);
```
parasyte, line 62: Why was this line removed?

agmcleod, line 62: @parasyte Just added back. I did changes to those files, and then merged with the latest 1.0.0-dev. I had merge conflicts with those files, looks like i deleted one line extra by mistake ><

PR #372, opened 2014-01-10
improved particle system performance

insidiator:
I tried the new particle system today and noticed, that it runs a bit slow, especially when using the additive mode.
After some profiling and tinkering around I managed to improve the performance a bit.
There is still much room for improvements, but I think it's a solid start.

The examples show the following fps improvements:

```
                         before       |      after
blending           | normal | lighter | normal | additive
Fire               |   46   |    26   |   60   |    40
Smoke              |   48   |    27   |   60   |    56
Fountain           |   16   |     7   |   60   |    38
Fireball           |   23   |    12   |   60   |    57
Rain               |   34   |    18   |   60   |    30
Cannon             |   60   |    60   |   60   |    60
Explosion Radial   |   35   |    16   |   60   |    25
Explosion Directed |   60   |    60   |   60   |    60
Waterfall          |   15   |    10   |   60   |    24
Multiples          |   21   |    14   |   60   |    33
```

The rain and waterfall example look a bit different than before, but I have not looked into the cause yet.

I think the matrix class could be helpful in managing the transformation of the individual particles.
Also I am planning to try a different approach to the additive mode, because it costs too much performance to use the globalCompositionOperation.

### DIFF
--- a/examples/platformer/js/entities/entities.js
+++ b/examples/platformer/js/entities/entities.js
@@ -16,8 +16,11 @@ game.PlayerEntity = me.ObjectEntity.extend({
 		this.setVelocity(3, 15);
 		this.setFriction(0.4,0);
 		
-		// update the hit box
-		this.updateColRect(20,32, -1,0);
+		// update the collision shape rect
+		var shape = this.getShape();
+		shape.pos.x = 20;
+		shape.resize(32, shape.height);
+
 		this.dying = false;
 		
 		this.mutipleJump = 1;
--- a/src/entity/entity.js
+++ b/src/entity/entity.js
@@ -375,23 +375,6 @@
 			}
 		},
 
-		/**
-		 * specify the size of the hit box for collision detection<br>
-		 * (allow to have a specific size for each object)<br>
-		 * @name updateColRect
-		 * @memberOf me.ObjectEntity
-		 * @function
-		 * @param {Number} x x offset (specify -1 to not change the width)
-		 * @param {Number} w width of the hit box
-		 * @param {Number} y y offset (specify -1 to not change the height)
-		 * @param {Number} h height of the hit box
-		 */
-		updateColRect : function(x, w, y, h) {
-			if (this.getShape().shapeType === "Rectangle") {
-				this.getShape().adjustSize(x, w, y, h);
-			}	
-		},
-
         /**
 		 * add a collision shape to this entity
 		 * @name addShape
--- a/src/shapes/rectangle.js
+++ b/src/shapes/rectangle.js
@@ -291,31 +291,6 @@
         },
 
         /**
-         * update the size of the collision rectangle<br>
-         * @deprecated use me.Rect.set()
-         * @name adjustSize
-         * @memberOf me.Rect
-         * @function
-         * @param {int} x x offset (specify -1 to not change the width)
-         * @param {int} w width of the hit box
-         * @param {int} y y offset (specify -1 to not change the height)
-         * @param {int} h height of the hit box
-         */
-        adjustSize : function(x, w, y, h) {
-            if (x !== -1) {
-                this.pos.x = x;
-                this.width = w;
-                this.hWidth = ~~(this.width / 2);
-            }
-                
-            if (y !== -1) {
-                this.pos.y = y;
-                this.height = h;
-                this.hHeight = ~~(this.height / 2);
-            }
-        },
-
-        /**
          *    
          * flip on X axis
          * usefull when used as collision box, in a non symetric way
--- a/src/system/timer.js
+++ b/src/system/timer.js
@@ -43,11 +43,10 @@
          * @ignore
          */
         var clearTimer = function (timerId) {
-            var i;
             for (var i = 0, len = timers.length; i < len; i++) {
                 if (timers[i].timerId === timerId) {
                     timers.splice( i, 1 );
-                    break
+                    break;
                 }
             }   
         };


src/particles/emitter.js
```diff
@@ -445,15 +462,85 @@
...
+
+            // Update particles if they are not dead yet
+            var particlesCount = this._particles.length;
+            var viewport = me.game.viewport;
+            var deadCount = 0;
+            for ( var i = 0; i < particlesCount; ++i) {
+                var particle = this._particles[i];
+                if(!particle.isDead) {
+                    particle.inViewport = (particle.pos.x < viewport.pos.x + viewport.width && 
+                                           viewport.pos.x < particle.pos.x + particle.width &&
```
obiot: http://melonjs.github.io/docs/me.Viewport.html#isVisible

src/particles/particle.js
```diff
@@ -67,11 +62,32 @@
...
+                    var sin = Math.sin(angle);
+                    var a = this.transform.a;
+                    var b = this.transform.b;
+                    var c = this.transform.c;
+                    var d = this.transform.d;
+                    this.transform.a = a * cos - b * sin;
+                    this.transform.b = a * sin + b * cos;
+                    this.transform.c = c * cos - d * sin;
+                    this.transform.d = c * sin + d * cos;
+                }
```
obiot: why not using the rotate function ?
https://github.com/melonjs/melonJS/blob/1.0.0-dev/src/math/matrix.js#L181

parasyte: My understanding is that these changes (many like it) are made with inlining optimizations in mind. Not ideal IMHO.

insidiator: If I use rotate here, the x and y position of the transform matrix would also be rotate.
I could call translate after I rotated the matrix, but I think a local rotation function would be neat to have here.

src/particles/emitter.js
insidiator: I know, I used that method before I inlined it, but there is something really bad going on with it in terms of performance.
The overlaps method of me.rect cost me around 5fps as you are using getters for left, right, top and bottom (https://bugzilla.mozilla.org/show_bug.cgi?id=626021).

src/particles/emitter.js
```diff
@@ -445,15 +486,62 @@
...
+
+        /**
+         * @ignore
+         */
+        draw : function(context, rect) {
+            if(this.children.length > 0) {
+                if(this.floating) {
+                    var viewport = me.game.viewport;
+                    this.transform.e = -viewport.screenX;
+                    this.transform.f = -viewport.screenY;
```
parasyte: That would be your issue. Don't do this transform. It breaks the `floating` property. `floating` objects are positioned relative to the screen itself. Like HUD items. Emitters must follow the same logic.

src/particles/particle.js
```diff
@@ -15,21 +14,22 @@
...
-            this.parent(emitter.pos.x + (Number.prototype.random(-emitter.varPos.x, emitter.varPos.x)),
-                        emitter.pos.y + (Number.prototype.random(-emitter.varPos.y, emitter.varPos.y)),
-                        emitter.image);
+            this.parent(new me.Vector2d(emitter.startPos.x + (Number.prototype.random(-emitter.varPos.x, emitter.varPos.x)),
+                        emitter.startPos.y + (Number.prototype.random(-emitter.varPos.y, emitter.varPos.y))),
+                        emitter.image.width, emitter.image.height);
 
             // Particle will always update
             this.alwaysUpdate = true;
+            this.floating = emitter.floating;
```
parasyte: Child objects already inherit this property.

src/particles/particle.js
```diff
@@ -15,21 +14,22 @@
...
         /**
          * @ignore
          */
         init: function(emitter) {
             // Call the parent constructor
-            this.parent(emitter.pos.x + (Number.prototype.random(-emitter.varPos.x, emitter.varPos.x)),
-                        emitter.pos.y + (Number.prototype.random(-emitter.varPos.y, emitter.varPos.y)),
-                        emitter.image);
+            this.parent(new me.Vector2d(emitter.startPos.x + (Number.prototype.random(-emitter.varPos.x, emitter.varPos.x)),
+                        emitter.startPos.y + (Number.prototype.random(-emitter.varPos.y, emitter.varPos.y))),
```
parasyte: Child objects are positioned relative to the container. So the emitter itself should be given an absolute position, and the particles should be given a position relative to it.

In other words, you should drop the `startPos` thing, and use just the position variance for the particle position vector.

src/particles/emitter.js
insidiator: this transform is what I had to add to get it to work. I know that the floating property should position the container relative to the screen, but for some reason it moves to 2_viewport.screenX, 2_viewport.screenY and thus moves out of view when I move the camera down and right.

src/particles/particle.js
insidiator: ah. that's why it is not moving as expected :)

src/particles/particle.js
insidiator: This is what I meant with a flag to prevent children of objectContainer to be transformed relative to the objectContainer. Particles should in fact not be relative to the emitter, because when I move the emitter around I don't want to move all particles together with it. They should stay in place and continue with their current path.

parasyte: In that case, the emitter should not be the container.

src/particles/emitter.js
```diff
@@ -348,6 +385,19 @@
...
+             * default value : 0 <br>
+             * @public
+             * @type Number
+             * @name framesToSkip
+             * @memberOf me.ParticleEmitter
+             */
+            this.framesToSkip = params.framesToSkip || 0;
+
+            // reset particle container values
+			me.game.world.addChild(this.container);
```
insidiator: I am missing a function which is called on addChild, like destroy is called on removeChild.
That way I could add the particle container to the world, when the emitter is added.

src/particles/emitter.js
```diff
@@ -65,33 +61,84 @@
...
             /**
-             * Z-order for particles <br>
-             * default value : 0
+             * Z-order for particles, value is forwarded to the particle container <br>
              * @type Number
              * @name z
              * @memberOf me.ParticleEmitter
              */
-            this.z = 0;
+            Object.defineProperty(this, "z", {
```
insidiator, line 95: Maybe this should be removed and only accessible directly on the container?
It's also quite useless without calling sort on the world, as the container is already added to the world when this is called.

src/particles/emitter.js
parasyte: If the emitter is added as a child of the ParticleContainer, this won't be an issue. But then you'll just end up duplicating `ObjectContainer` functionality, because there are a lot of concepts that you need to take into consideration in regard to scene graphs, and a lot of things that you can't safely optimize out.

Issue #157, opened 2013-02-11
add support for Web Audio API


ghost:
As discussed in Ticket #110 

With fallback to HTML5 audio if not supported 

**To be implemented**:
- Remove built in audio support detection, utilize the me.audio.init or ext value from resources for what URLs are available.
- Re-implement something like channels.

PR #376, opened 2014-01-14
Feature/howler audio

agmcleod:
#157


src/audio/audio.js
```diff
@@ -265,10 +256,13 @@
 				}
 				sync_loading = true;
 			}
-
+			var urls = [];
+			for(var i = 0; i < this.audioFormats.length; i++) {
+				urls.push(sound.src + sound.name + "." + this.audioFormats[i] + me.loader.nocache);
+			}
```
parasyte, line 262: I like the `Array.prototype.map` function, better. ;)

agmcleod, line 262: I'm old fashioned i guess ;). Map is a pretty function, i use it in ruby quite a bit. I also know it tends to be a tad slower than the for counter party.

parasyte, line 262: For an upper bound of 5 array elements? ;)

agmcleod, line 262: Hello world has never run so slowly.

To be honest, I didn't even think of it that much. I just kinda wrote the array to build it haha.

PR #472, opened 2014-03-26
New tests environment

ellisonleao:
Using the same text as #469 but now merging to the right branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 melonJS
 =======
 [![Build
-Status](https://travis-ci.org/melonJS/melonJS.svg?branch=master)](https://travis-ci.org/melonJS/melonJS)
+Status](https://travis-ci.org/melonJS/melonJS.svg?branch=1.1.0-dev)](https://travis-ci.org/melonJS/melonJS)
 [![GitHub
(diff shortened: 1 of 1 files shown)


src/system/device.js
```diff
@@ -112,27 +112,30 @@
...
-                        }
-                    } else {
-                        if (me.sys.stopOnBlur) {
-                            me.state.restart(true);
-                        }
-                        if (me.sys.resumeOnFocus) {
-                            me.state.resume(true);
+                 
+            // register on the event if supported
+            if (typeof (visibilityChange) === "string") {
```
agmcleod, line 117: @ellisonleao What's the purpose behind the conditional change here?

obiot, line 117: CocoonJS compatibility, as `visibilityChange` remains undefined in that case !

(this is from master though)

agmcleod, line 117: Ah okay, was just surprised to see changes to the actual logic in the PR :)

ellisonleao, line 117: Yeah that was not me, this came with the master branch merge. @agmcleod

agmcleod, line 117: ah okay. Interesting that we're seeing the diff.

PR #502, opened 2014-05-17
Fixes in ParticleEmitter and New Keys Bidings in Input

aaschmitz:
I send the pull request for the version master, but this is for 1.0.x ;) Now, the spaces is OK.

### DIFF
--- a/src/input/keyboard.js
+++ b/src/input/keyboard.js
@@ -119,26 +119,34 @@
      */
 
     /**
-     * list of mappable keys :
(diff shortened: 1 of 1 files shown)


src/input/keyboard.js
```diff
@@ -119,26 +119,34 @@
...
      * @enum {number}
      * @name KEY
      * @memberOf me.input
      */
     obj.KEY = {
+        "BACKSPACE" : 8,
         "TAB" : 9,
         "ENTER": 13,
         "SHIFT" : 16,
         "CTRL": 17,
```
parasyte, line 134: This line, for example. You've added a single space before colons on other lines in this area of the code. But missed this line.

PR #524, opened 2014-06-24
Inlined some method calls in me.TMXOrthogonalRenderer.drawTileLayer

johnrayner:
A couple of method calls in drawTileLayer were creating garbage unnecessarily, i.e. `me.Vector2d` objects which were immediately released.  Inlining these methods gives us slightly less elegant code but saves us some cycles on garbage collection.

### DIFF
--- a/src/level/TMXRenderer.js
+++ b/src/level/TMXRenderer.js
@@ -38,6 +38,7 @@
 
         /**
          * return the tile position corresponding to the specified pixel
+         * to avoid memory leaks, be sure to me.pool.push the return value of this function when you are done with it
          * @ignore
          */
         pixelToTileCoords : function (x, y) {
@@ -48,6 +49,7 @@
 
         /**
          * return the pixel position corresponding of the specified tile
+         * to avoid memory leaks, be sure to me.pool.push the return value of this function when you are done with it
          * @ignore
          */
         tileToPixelCoords : function (x, y) {
@@ -148,6 +150,7 @@
 
         /**
          * return the tile position corresponding to the specified pixel
+         * to avoid memory leaks, be sure to me.pool.push the return value of this function when you are done with it
          * @ignore
          */
         pixelToTileCoords : function (x, y) {
@@ -161,6 +164,7 @@
 
         /**
          * return the pixel position corresponding of the specified tile
+         * to avoid memory leaks, be sure to me.pool.push the return value of this function when you are done with it
          * @ignore
          */
         tileToPixelCoords : function (x, y) {
@@ -297,4 +301,4 @@
 			me.pool.push(startPos);
         }
     });
-})();
+})()


src/renderable/container.js
```diff
@@ -19,7 +19,9 @@
      * A global "translation context" for nested ObjectContainers
      * @ignore
      */
-    var globalTranslation = new me.Rect(new me.Vector2d(), 0, 0);
+	//var zeroPoint = me.pool.pull("me.Vector2d", 0, 0);
+    var globalTranslation = new me.Rect(new me.Vector2d(0, 0), 0, 0);
+	//me.pool.push(zeroPoint);
```
parasyte: I don't understand this change. Did you forget to fix it up?

src/level/TMXLayer.js
```diff
@@ -19,7 +19,9 @@
         // constructor
         init: function (name, color, z) {
             // parent constructor
-            this._super(me.Renderable, "init", [new me.Vector2d(0, 0), Infinity, Infinity]);
+			var zeroPoint = me.pool.pull("me.Vector2d", 0, 0);
+            this._super(me.Renderable, "init", [zeroPoint, Infinity, Infinity]);
+			me.pool.push(zeroPoint);
```
parasyte: melonJS source code uses four spaces for indentation. Please normalize your tabs to four spaces.

src/level/TMXRenderer.js
```diff
@@ -249,7 +255,7 @@
             var shifted = inUpperHalf ^ inLeftHalf;
 
             // initialize the columItr vector
-            var columnItr = rowItr.clone();
+            var columnItr = me.pool.pull("me.Vector2d", rowItr.x, rowItr.y);
```
parasyte: This would be better if the `clone` method returned a vector from the pool instead of `new`...

parasyte: Or rather a new method, to retain backward compatibility. `me.Vector2D.poolClone()` ?

src/system/pooling.js
```diff
@@ -187,4 +188,4 @@
         // return our object
         return api;
     })();
-})();
+})();
```
parasyte: I kind of like having the blank line at the end of the files. No need to remove it, anyway.

src/renderable/container.js
johnrayner: In the unit tests, this code executes before the me.pool.init method has been called.  The impact of this particular line is also pretty low, since AFAICS it's executed only once as Melon initializes.  But I shouldn't have left the commented code in place.

src/level/TMXLayer.js
```diff
@@ -19,7 +19,9 @@
         // constructor
         init: function (name, color, z) {
             // parent constructor
-            this._super(me.Renderable, "init", [new me.Vector2d(0, 0), Infinity, Infinity]);
+            var zeroPoint = me.pool.pull("me.Vector2d", 0, 0);
+            this._super(me.Renderable, "init", [zeroPoint, Infinity, Infinity]);
+            me.pool.push(zeroPoint);
```
parasyte, line 24: Are these vectors used later? Certainly they continue to be referenced by this layer, so it's not exactly safe to push them back into the pool until the layer itself is removed from the scene and destroyed... Same with the other changes in this file. I guess there's a better way to do this, or just leave it as it was (no pool).

src/level/TMXObjectGroup.js
```diff
@@ -317,18 +317,26 @@
...
          */
         getShape : function () {
+			var zeroPoint = me.pool.pull("me.Vector2d", 0, 0);
+			
             // add an ellipse shape
             if (this.isEllipse === true) {
-                return new me.Ellipse(new me.Vector2d(0, 0), this.width, this.height);
+                var ellipse = new me.Ellipse(zeroPoint, this.width, this.height);
+				me.pool.push(zeroPoint);
+				return ellipse;
```
parasyte, line 326: This set of changes is definitely not safe, because it returns a reference to the shape, where code may then attempt to manipulate the position vector after it has been pushed into the pool.

src/level/TMXRenderer.js
```diff
@@ -41,8 +41,9 @@
          * @ignore
          */
         pixelToTileCoords : function (x, y) {
-            return new me.Vector2d(x / this.tilewidth,
-                                   y / this.tileheight);
+            return me.pool.pull("me.Vector2d",
+								x / this.tilewidth,
+                                y / this.tileheight);
```
parasyte, line 47: Are you sure you are pushing these vectors (pixelToTile and tileToPixel) back into the pool when you are done with them? I did a quick page-search for these method names and did not find any references that would suggest that is taken care of. This will cause a "memory leak" if it is not handled properly.

Second, if these methods may ever be called by user code (in this case they are ignored by documentation) then this behavior would need to be well-documented. Or better, avoided altogether.

parasyte, line 47: Ok! Looks like pushing these vectors is already done, actually. I didn't have the context lines expanded in the patch to see them. The note about documentation could be important though.

src/level/TMXObjectGroup.js
johnrayner, line 326: Shapes don't hold a reference to the position vector which is passed to them.  They universally create their own position vector and then call `setV` on it, which simply copies `x` and `y` values.

src/level/TMXLayer.js
johnrayner, line 24: I've looked through the base class code here, and `me.Renderable` passes the position vector into `me.Rect`, which creates a new `me.Vector2d` and then copies values from it.  So I'm pretty sure this change is safe, but perhaps we don't want to couple to base class implementation so tightly.

parasyte, line 24: Wow! That's a pretty heavy operation. :unamused: Perhaps we could do that instantiation better internally...

agmcleod, line 24: I almost wonder if we should refactor the constructors to just accept x & y instead of a vector object. Let Rect handle the assignment. Would keep it consistent with animation sheet, sprite sheet and entity.

src/level/TMXRenderer.js
johnrayner, line 47: I'll add in some documentation, but my reasoning about user code calling these methods went like this.  The worst case afaics is that user's forget to push the vector.  This would tend to exhaust the vector pool prematurely and would add to GC pressure.  While this isn't great, I can't see it would leak memory or cause any unexpected side-effects.

parasyte, line 47: The reason it leaks is because the pool grows without bounds, and will never automatically release memory. So it could be troublesome.

src/level/TMXLayer.js
obiot, line 24: i actually thought about this so many times, but never had the "balls" to do it (due to too many changes)

agmcleod, line 24: If you don't completely break your library now and then, where's the fun? :)

Issue #508, opened 2014-05-24
Better release strategy (with bower)


parasyte:
@agmcleod @obiot I have been investigating [bower](http://bower.io/) more closely, which was discussed a bit in #227.

My understanding of the bower package manager ecosystem is still vague, at best. But it does appear to be useful for the distribution of *pre-built* melonJS releases -- something we are NOT doing with it yet.

In my experiments, I installed jQuery using `bower install jquery` in an empty directory. It creates a subdirectory called `bower_components` (synonymous to `node_modules` with `npm`) and jQuery is installed there. What's interesting to note is that in the `jquery` subdirectory, there's the full source code, plus the pre-built js files in a `dist` directory. And the bower manifest points to one of these dist files: https://github.com/jquery/jquery/blob/master/bower.json#L4

But if you look at the jQuery repo, you'll see that there is no dist directory, and in fact, it is ignored in `.gitignore`! But if you check out their latest tag, you will indeed find the dist directory: https://github.com/jquery/jquery/tree/2.1.1/dist This means the `2.1.1` tag is not in any branch. So jQuery's dist/bower register step must detach the branch, commit the dist files, tag the commit, and push the detached tag.

This is the flow I tested in a new repo, and it works really well: https://github.com/parasyte/hork The build directory is purposefully empty (just a placeholder) And I have made three tags, each with a different build file that never touched a branch, with all tags having histories from different points in the master branch. Here's how we might do this with melonJS:

```
grunt
git checkout --detach
git add -f build/melonjs-1.0.1-min.js
git add -f build/melonjs-1.0.1.js
git commit -m "Release 1.0.1"
git tag 1.0.1
git push origin 1.0.1
```

What we're doing here is building the dist files with grunt, then detaching from the branch and committing the dist files in the detached state. This creates the commit sha1 in git history, but the commit isn't really linked anywhere. To link it with something, a tag is created (which integrates with the github release mechanism) and the tag is pushed to remote (github) to create the release.

And of course, all of these steps can be automated to pull the version number from package.json.

With all of that done, we should be able to `bower install melonJS` and get the dist files (plus source code) installed into our project when starting a new game. No need to build melonJS, no need to install any build dependencies, and finally the `npm install` task does not need to run bower; they can live entirely independent of one another, the way it should be.

----

[...]

PR #528, opened 2014-07-08
Create first version of release task [#508]

ellisonleao:


tasks/release.js
```diff
@@ -0,0 +1,80 @@
...
+            ];
+            // check the build files from the actual version
+            // and add the js files to be commited
+            var stringFiles = '';
+            for (var i = 0; i < filenames.length; i++) {
+                if (grunt.file.exists(filenames[i])) {
+                    stringFiles += filenames[i] + ' ';
+                }
+            }
+            run('git add -f ' + stringFiles, 'Adding build files');
```
agmcleod: the force option worries me slightly. Could this include files we have ignored?

ellisonleao: No. this only includes files from the `filenames` array.

tasks/release.js
```diff
@@ -0,0 +1,80 @@
+'use strict';
+var path = require('path');
+var Q = require('q');
+var shell = require('shelljs');
```
parasyte: `q` and `shelljs` need to be added to package.json `devDependencies`.

tasks/release.js
```diff
@@ -0,0 +1,80 @@
...
+
+module.exports = function(grunt) {
+    grunt.registerTask('dorelease', 'MelonJS Release', function() {
```
parasyte: This task (and all tasks, actually!) should be getting validated by jshint. It will catch style issues like the single-quoted strings here. It also finds things like the missing semi-colon on line 23. :smiley:

tasks/release.js
```diff
@@ -0,0 +1,80 @@
...
+                path.join(buildPath, 'melonjs-' + version + '.js'),
+                path.join(buildPath, 'melonjs-' + version + '-min.js')
+            ];
+            // check the build files from the actual version
+            // and add the js files to be commited
+            var stringFiles = '';
+            for (var i = 0; i < filenames.length; i++) {
+                if (grunt.file.exists(filenames[i])) {
+                    stringFiles += filenames[i] + ' ';
+                }
```
parasyte, line 62: If this check fails, I want the task to fail. :wink:

tasks/release.js
```diff
@@ -0,0 +1,80 @@
...
+        function commit() {
+            run('git commit -am "Release '+ version  +'"', 'Commiting release');
+        }
+
+        function tag() {
+            run('git tag ' + version, 'Tagging new version');
+        }
+
+        function push() {
+            run('git push origin master', 'Pushing to master branch');
```
parasyte: This is not correct. It needs to push to the tag on origin, NOT `master`!

``` javascript
"git push origin " + config.version
```

tasks/release.js
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        function back() {
```
parasyte: Nitpick! This function name is not descriptive enough. How about `rollback`?

tasks/release.js
```diff
@@ -0,0 +1,80 @@
...
+        function checkout() {
+            run('git checkout --detach', 'Detaching from current tree');
+        }
+
+        function add() {
+            grunt.log.oklns('ACTUAL VERSION ==> ' + config.version);
+            grunt.log.oklns('BUILD FILES');
+            var filenames = [
+                path.join(buildPath, 'melonjs-' + version + '.js'),
+                path.join(buildPath, 'melonjs-' + version + '-min.js')
```
parasyte: For these filenames, you should be able to use `config.path.main` and `config.path.min`.

tasks/release.js
```diff
@@ -0,0 +1,80 @@
...
+          .then(back)
+          .catch(function(msg){
+            grunt.fail.warn(msg || 'release failed')
+          })
+          .finally(done);
+
+        function run(cmd, msg){
+            var deferred = Q.defer();
+            grunt.verbose.writeln('Running: ' + cmd);
+            var success = shell.exec(cmd, {silent:true}).code === 0;
```
parasyte: The value of `silent` here should be passed as the value of `!grunt.option("verbose")`

tasks/release.js
```diff
@@ -0,0 +1,83 @@
...
+            var filenames = [
+                path.join(buildPath, "melonjs-" + version + ".js"),
+                path.join(buildPath, "melonjs-" + version + "-min.js")
+            ];
+            // check the build files from the actual version
+            // and add the js files to be commited
+            var stringFiles = "";
+            for (var i = 0; i < filenames.length; i++) {
+                if (!grunt.file.exists(filenames[i])) {
+                    grunt.fail.fatal("Missing file " + filenames[i] + " is required");
```
parasyte, line 61: Can you move the "rollback" command to something like a "finally" block? Where it will be executed even if this line aborts the grunt task. That will ensure the git repo isn't left in a broken state.

ellisonleao, line 61: Perhaps if i add it into the `catch` statement , just before the `grunt.fail.warn` call, do you think that should do it?

tasks/release.js
```diff
@@ -0,0 +1,83 @@
...
+        try {
+            checkout();
+            add();
+            commit();
+            tag();
+            push();
+        } catch (err) {
+            grunt.fail.warn(err || "release failed");
+        }
+        rollback();
```
parasyte: This should probably be:

``` javascript
finally {
    rollback();
}
```

package.json
```diff
@@ -33,7 +33,9 @@
     "grunt-contrib-jshint": "~0.8.0",
     "grunt-contrib-uglify": "~0.2.7",
     "grunt-jsdoc": "~0.5.1",
-    "grunt-replace": "~0.7.7"
+    "grunt-replace": "~0.7.7",
+    "q": "~1.0.1",
```
parasyte: No longer needs `q`

tasks/release.js
```diff
@@ -0,0 +1,83 @@
+"use strict";
+var path = require("path");
+var Q = require("q");
```
parasyte: Remove `q` and the deferred promise in the `run()` function below.

tasks/release.js
```diff
@@ -0,0 +1,83 @@
...
+        function tag() {
+            run("git tag " + version, "Tagging new version");
+        }
+
+        function push() {
+            run("git push origin " + version, "Pushing to new version branch");
+        }
+
+        function rollback() {
+            run("git checkout master", "Getting back to master branch");
```
parasyte: If the release task fails between `add` and `commit` for some reason, this rollback command will fail, because you can't do a checkout if you have staged changes (from `add`). A simple workaround is to `git reset` before the checkout:

``` javascript
run("git reset --hard", "Resetting staged changes");
```

ellisonleao: or i can add a `-f` on the checkout. `git checkout -f`

tasks/release.js
ellisonleao: i think the `finnaly` is kinda unnecessary in this point form my perspective, since the `rollback()` function will be executed by the end everytime. Don't you agree?

parasyte: Semantically you are likely correct. However, the use of `finally` is a better practice, as code inside the `finally` block will execute even in the unlikely event that an exception is raised in the `catch` block, like somewhere deep in the `grunt.fail.warn` call... Even if the exception is something you wouldn't normally expect, like a keyboard interrupt (Ctrl+C).

It's just a way to ensure the "cleanup" code is _always_ run, even when unforeseen troubles arise.

ellisonleao: I see. Let me add then. thanks!

tasks/release.js
```diff
@@ -0,0 +1,97 @@
...
+            }
+        }
+
+        function checkout() {
+            var symbolicRef = shell.exec("git symbolic-ref HEAD", shellOpts).output;
+            if (symbolicRef) {
+                var splitted = symbolicRef.split('/');
+                // symbolic ref must contain 3 steps
+                // e.g ref/heads/master
+                if (splitted.length !== 3) {
```
parasyte: It is possible (and recommended, in some cases!) to name branches with a forward slash. This check would be better like `>= 3`

Then the `currBranch` assignment below would be like:

``` javascript
currBranch = splitted.slice(2).join("/");
```

tasks/release.js
```diff
@@ -0,0 +1,97 @@
...
+                    rollback();
+                    grunt.fail.fatal("Missing file " + filenames[i] + " is required");
+                }
+                stringFiles += filenames[i] + " ";
+            }
+            run("git add -f " + stringFiles, "Adding build files");
+        }
+
+        function commit() {
+            run("git commit -am 'Release " + version + " '", "Commiting release");
```
parasyte, line 69: Nitpick: There is a stray space in the command, just after the version.

tasks/release.js
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        function tag() {
+            run("git tag " + version, "Tagging new version");
```
parasyte, line 73: Might as well use annotated tags so we can tell who created the tag and when! Add `-a` and `-m` flags to the git command here. See: http://git-scm.com/book/en/Git-Basics-Tagging#Annotated-Tags

PR #535, opened 2014-07-18
Remove unneeded code from src/lang/string.js

gravypod:
Save the extra trim.

### DIFF
--- a/src/lang/number.js
+++ b/src/lang/number.js
@@ -74,12 +74,9 @@ Number.prototype.round = function () {
  * @alias toHex
  * @return {string} converted hexadecimal value
  */
-Number.prototype.toHex = (function () {
-    var hexString = "0123456789ABCDEF";
-    return function () {
-        return hexString.charAt((this - (this % 16)) >> 4) + hexString.charAt(this % 16);
-    };
-})();
+Number.prototype.toHex = function () {
+    return "0123456789ABCDEF".charAt((this - (this % 16)) >> 4) + "0123456789ABCDEF".charAt(this % 16);
+};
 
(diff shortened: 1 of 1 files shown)


src/lang/string.js
```diff
@@ -64,10 +64,11 @@ String.prototype.isNumeric = function () {
  * @return {boolean} true if the string is either true or false
  */
 String.prototype.isBoolean = function () {
-    return (
-        this !== null &&
-        ("true" === this.trim() || "false" === this.trim())
-    );
+    if (this !== null) {
```
parasyte: If `this` is a boolean, this condition will cause the function to always return `false`. Change it to `===`.

gravypod: Wouldn't you only need to change "("true" === trimmed) || ("false" === trimmed)" to "("true" == trimmed) || ("false" == trimmed)". And how would a Boolean inherit String.prototype?

parasyte: Right, I missed that entirely. `this` can never be null, unless code is doing something silly like `String.prototype.isBoolean.call(null)`. So it's really an unnecessary safety check.

The triple-equals is good from an optimization perspective since it doesn't do any type coercion.

gravypod: If you are going to remove the `=== null`s , there are a few extra things I could change before you merge into master. Do you have an email or IRC server you go on? I am on irc.esper.net and you can email at {myusername}@{myusername}.com

parasyte: We have an IRC channel on Freenode #melonJS. You may also contact me at jay@kodewerx.org

PR #537, opened 2014-07-22
getTile returns wrong tile with isometric maps

juhanapaavola:
Because of orientation, width/height cannot be used directly to get
tile.

### DIFF
--- a/src/level/TMXLayer.js
+++ b/src/level/TMXLayer.js
@@ -489,7 +489,24 @@
          * @return {me.Tile} Tile Object
          */
         getTile : function (x, y) {
-            return this.layerData[~~(x / this.tilewidth)][~~(y / this.tileheight)];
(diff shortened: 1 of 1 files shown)


src/level/TMXLayer.js
```diff
@@ -489,7 +489,24 @@
...
+            if (this.orientation === "isometric") {
+                var tmpx = x;
+                var tmpy = y;
+                
+                tmpx -= this.rows * this.tilewidth / 2;
+                var tilex = Math.round(tmpx / this.tilewidth);
+                var tiley = Math.round(tmpy / this.tileheight);
+                tx = tiley + tilex;
+                ty = tiley - tilex;
+
```
obiot, line 504: this all look like it could be optimized a bit further, see here :
https://github.com/melonjs/melonJS/blob/master/src/level/TMXRenderer.js#L148

Issue #498, opened 2014-05-12
Improve `me.Font` and `me.BitmapFont` classes by using `me.Color` class for color definition/management 


obiot:
see #486

PR #615, opened 2014-12-06
Updated me.Font to work with me.Color

xorinzor:
Updated me.Font to work with me.Color

only the tween at font_test example acts a bit weird

This is a (possible) solution for https://github.com/melonjs/melonJS/issues/498

### DIFF
--- a/examples/devicetest/js/screens/play.js
+++ b/examples/devicetest/js/screens/play.js
@@ -14,7 +14,8 @@ game.PlayScreen = me.ScreenObject.extend({
         me.game.world.addChild(new (me.Renderable.extend({
             init: function() {
                 this._super(me.Renderable, 'init', [0, 0, 100, 200]);
-                this.font = new me.Font('arial', '24px', new me.Color(255, 255, 255));
+                this.fontColor = me.pool.pull("me.Color", 255, 255, 255);
+                this.font = new me.Font('arial', '24px', this.fontColor);
             },
             update: function() {
                 return true;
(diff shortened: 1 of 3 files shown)


examples/font_test/main.js
```diff
@@ -101,16 +101,16 @@ var FontTest = me.Renderable.extend ({
...
-        this.font.setFont('Arial', 16, this.color.toHex());
+        this.font.setFont('Arial', 16, this.color);
         var baseline = 200;
 
         // Draw the baseline
         context.beginPath();
         context.moveTo(0, baseline + 0.5);
         context.lineTo(me.video.renderer.getWidth(), baseline + 0.5);
-        context.strokeStyle = "red";
+        context.strokeStyle = new me.Color().parseCSS("red");
```
parasyte: I don't like this use of `parseCSS` ... why do that when "red" is perfectly acceptable?

examples/devicetest/js/screens/play.js
```diff
@@ -14,7 +14,7 @@ game.PlayScreen = me.ScreenObject.extend({
         me.game.world.addChild(new (me.Renderable.extend({
             init: function() {
                 this._super(me.Renderable, 'init', [0, 0, 100, 200]);
-                this.font = new me.Font('arial', '24px', '#fff');
+                this.font = new me.Font('arial', '24px', new me.Color(255, 255, 255));
```
parasyte: We can use the pool to create temporary colors:

``` javascript
me.pool.pull("me.Color", 255, 255, 255)
```

Just remember to push the color back into the pool when you're done using it!

src/font/font.js
```diff
@@ -36,19 +36,19 @@
              * defines the color used to draw the font.<br>
              * Default value : "#000000"
              * @public
-             * @type String
+             * @type me.Color
              * @name me.Font#fillStyle
              */
-            this.fillStyle = fillStyle || "#000000";
+            this.fillStyle = (fillStyle instanceof me.Color) ? fillStyle : new me.Color(0, 0, 0);
```
parasyte: TBH, `fillStyle` is not documented as optional. Either we should change the documentation, or always assume it will be given a valid color.

Issue #258, opened 2013-08-16
add emulation support for `mouseenter` and `mouseleave` events


obiot:
this was already discussed and requested with ticket #115, but since these 2 ones are the only one remaining, i'm creating this one in favor of the other one (and will close it).

obiot:
we can close this one then!

PR #629, opened 2015-01-03
#258

TheManuz:
I've finally been able to modify the event system.
##### pointer.js
I needed to modify the structure that keeps the reference of handlers for many reasons:
* when a _pointermove_ event goes out or inside the bounds of a rect, when need to fire _pointerleave_ or _pointerenter_ events. But for this to happen we need to process _pointermove_ events even if the object is not listening for them, because it may be listening for _pointerleave_ or _pointerenter_ instead.
* we needed a way to access to all the callbacks tied to a specific __rect__, in order to dispatch the right event
* we needed every callback to being able to access the pointerId, in order to know if the object is already clicked or not and by what touch

The "evtHandlers" structure is now a dictionary where the objects gets saved by their GUID (they are assigned one if they don't have it). For each object we save:
```javascript
evtHandlers[rect.GUID] = {
	rect : rect,		// the rect reference
	callbacks : {},		// a callbacks dictionary (to be accessed with the event names)
	pointerId : null,	// needed at this level, accessible for every callback
};
```
The callback dictionary contains, for each event name, an array containing callbacks.
Every time a pointerEventCallback gets registered, it's pushed into this array (this is pretty much how it was before, just a level deeper into the structure):
```javascript
evtHandlers[rect.GUID].callbacks[eventType].push(callback);
};
```
##### GUI.js
I also modified the GUI_Object to use the new event system.
I tested it on Google Chrome and on CocoonJS on Android (it finally handles correctly when a touch slides off the screen: it fires a "pointerCancel" and the button goes unpressed. It also behaved good when sliding a finger between 2 buttons: fires "pointerEnter" and "pointerLeave", the buttons get pressed and unpressed correctly).

It may need other tests, though.

### DIFF
--- a/src/input/pointer.js
+++ b/src/input/pointer.js
@@ -144,12 +144,6 @@
     var viewportOffset = new me.Vector2d();
 
     /**
-     * Array of object containing changed touch information (iOS event model)
-     * @ignore
-     */
-    var changedTouches = [];
-
-    /**
      * cache value for the offset of the canvas position within the page
      * @ignore
      */
@@ -174,7 +168,7 @@
     function enablePointerEvent() {
         if (!pointerInitialized) {
             // initialize mouse pos (0,0)
-            changedTouches.push({ x: 0, y: 0 });
+            // changedTouches.push({ x: 0, y: 0 });
             obj.mouse.pos = new me.Vector2d(0, 0);
             // get relative canvas position in the page
             obj._offset = me.video.getPos();
@@ -241,7 +235,7 @@
      * propagate events to registered objects
      * @ignore
      */
-    function dispatchEvent(e) {
+    function dispatchEvent(e, changedTouches) {
         var handled = false;
         var handlers = evtHandlers[e.type];
 
@@ -314,8 +308,8 @@
     function updateCoordFromEvent(event) {
         var local;
 
-        // reset the touch array cache
-        changedTouches.length = 0;
+        // touch array cache
+        var changedTouches = [];
 
         // PointerEvent or standard Mouse event
         if (!event.touches) {
@@ -334,13 +328,14 @@
         }
         // if event.isPrimary is defined and false, return
         if (event.isPrimary === false) {
-            return;
+            return changedTouches;
         }
         // Else use the first entry to simulate mouse event
         obj.mouse.pos.set(
             changedTouches[0].x,
             changedTouches[0].y
         );
+		return changedTouches;
     }
 
 
@@ -364,8 +359,9 @@
                 // Webkit also support wheelDeltaX
                 e.wheelDeltaX && (_event.deltaX = - 1 / 40 * e.wheelDeltaX);
             }
+			var changedTouches = updateCoordFromEvent(_event);
             // dispatch mouse event to registered object
-            if (dispatchEvent(_event)) {
+            if (dispatchEvent(_event, changedTouches)) {
                 // prevent default action
                 return obj._preventDefault(e);
             }
@@ -380,9 +376,9 @@
      */
     function onMoveEvent(e) {
         // update position
-        updateCoordFromEvent(e);
+        var changedTouches = updateCoordFromEvent(e);
         // dispatch mouse event to registered object
-        if (dispatchEvent(e)) {
+        if (dispatchEvent(e, changedTouches)) {
             // prevent default action
             return obj._preventDefault(e);
         }
@@ -395,10 +391,10 @@
      */
     function onPointerEvent(e) {
         // update the pointer position
-        updateCoordFromEvent(e);
+        var changedTouches = updateCoordFromEvent(e);
 
         // dispatch event to registered objects
-        if (dispatchEvent(e)) {
+        if (dispatchEvent(e, changedTouches)) {
             // prevent default action
             return obj._preventDefault(e);
         }
--- a/src/input/pointer.js
+++ b/src/input/pointer.js
@@ -240,10 +240,10 @@
     function dispatchEvent(e, changedTouches) {
         var handled = false;
         var handlers = evtHandlers[e.type];
-
+		var eventType = activeEventList.indexOf(e.type);
         // Convert touchcancel -> touchend, and pointercancel -> pointerup
         if (!handlers) {
-            if (activeEventList.indexOf(e.type) === POINTER_CANCEL) {
+            if (eventType === POINTER_CANCEL) {
                 handlers = evtHandlers[activeEventList[POINTER_UP]];
             } else {
                 handlers = evtHandlers[e.type];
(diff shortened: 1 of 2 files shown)


parasyte:
@TheManuz This is awesome! I'll start going through the changes, and will let you know if I spot anything that needs to be better. Standard code review procedure. :)

TheManuz:
I'm sure the review is really needed, but i hope there won't be big problems! :D

src/input/pointer.js
```diff
@@ -241,68 +237,141 @@
...
+											handled = true;
+											break;
+										}
+									}
+								}
+							}
+							break;
+					}
+				}
+			}
```
parasyte: Our coding conventions require 4 spaces instead of tab characters. Please convert all tabs to spaces in both files that were modified. That will help align the code properly on github, and make the rest of it easier to review.

Also, we have a file in the project root directory for editor configuration that will help for these kinds of settings: https://github.com/melonjs/melonJS/blob/master/.editorconfig

src/input/pointer.js
```diff
@@ -314,8 +383,8 @@
     function updateCoordFromEvent(event) {
         var local;
 
-        // reset the touch array cache
-        changedTouches.length = 0;
+        // touch array cache
+        var changedTouches = [];
```
obiot: I would rather keep this one declared "globally" as before to avoid creating ana array object each time the function is called (this is why it was declared global before)

src/input/pointer.js
```diff
@@ -560,7 +631,7 @@
      * // register on the 'pointerdown' event
      * me.input.registerPointerEvent('pointerdown', this, this.pointerDown.bind(this));
      */
-    obj.registerPointerEvent = function (eventType, rect, callback, floating) {
```
obiot, line 563: so if you removed the floating property, how does it work now ?

src/input/pointer.js
```diff
@@ -572,25 +643,28 @@
...
+        if (!rect.GUID) {
+            rect.GUID = me.utils.createGUID();
+        }
+        
         // register the event
-        if (!evtHandlers[eventType]) {
-            evtHandlers[eventType] = [];
+        if (!evtHandlers[rect.GUID]) {
+            evtHandlers[rect.GUID] = {
+                rect : rect,
```
obiot, line 659: actually we should maybe change this part, change the function signature and replace the `rect` parameter by any object implementing the `getBounds()` function (and stores the object reference instead of the rect) and of course change any use of rect by a call to the object `getBounds()` function. 

Then If a `me.Rect` is provided, the behavior will probably be the same, but then if we pass an entity, it will allow to provide dynamic mouse detection for entities with changing shapes.

obiot:
did you also test on iOS devices ? these are the tricky part with the standard (pointerEvent) emulation. Although older Android devices (and Cocoonjs) also use it so that should also work.

But globally the device/browser to test are :
* W7/8 - IE10 and IE11
* WIN - Chrome
* OSX - Chrome /Safari
* iOS 7/8 (webview and cocoonJS)
* Android (native browser and cocoonJS)

src/input/pointer.js
TheManuz: You're right. In a previous implementation, when a move event needed to be re-dispatched as a enter/leave event, the global changedTouches caused many problems. I think I can switch it again to the global one, it should work with the current implementation.

src/input/pointer.js
TheManuz, line 563: Since it saves a reference to the rect, it check if the rect is floating in the first place.

src/input/pointer.js
TheManuz, line 659: I totally agree on this.

TheManuz:
Ehm, sadly I'm working on an old laptop with Windows XP, so i cannot test on OSX. Also I've only got an Android smartphone, so i cannot test on iOS!
When I'll submit the new changes, I hope someone can test them on other platforms.

parasyte:
No worries on compatibility testing. I think we can do that here before merging. :)

src/renderable/GUI.js
```diff
@@ -146,16 +160,6 @@
...
-         * function callback for the pointerup event
-         * @ignore
-         */
-        release : function (event) {
-            this.released = true;
-            me.timer.clearTimeout(this.holdTimeout);
-            return this.onRelease(event);
-        },
-
-        /**
```
parasyte, line 158: nit: Why was this moved?

obiot:
indeed the compatibility list was just a global reminder for all of us, based on my experience when working on this.

src/renderable/GUI.js
TheManuz, line 158: Right question!
When i was working on the GUI_Object, i also added **moved**, **entered** and **leaved** function, and i moved all of them together with **clicked** and **release** to the top.
On the bottom i brought together the **onClick**, **onRelease** and the (now missing) **onMove**.
After some fiddling i removed unnecessary functions, but the order was left changed.

obiot:
Hi @TheManuz, sorry i have not been able to follow on this one recently, so any update her, what is the status of this one, especially in terms of testing ?

src/input/pointer.js
```diff
@@ -243,63 +245,136 @@
...
-            // get the current screen to world offset
-            me.game.viewport.localToWorld(0, 0, viewportOffset);
-            for (var t = 0, l = changedTouches.length; t < l; t++) {
-                // Do not fire older events
-                if (typeof(e.timeStamp) !== "undefined") {
-                    if (e.timeStamp < lastTimeStamp) {
-                        continue;
+        
+        for (var guid in evtHandlers) {
+            if (evtHandlers.hasOwnProperty(guid)) {
```
parasyte, line 250: This for-in-hasOwnProperty loop should be changed to:

``` javascript
Object.keys(evtHandlers).forEach(function (guid) {
    // ...
});
```

TheManuz, line 250: @parasyte: doesn't creating an anonymous function everytime adds overhead?
@obiot: testing has been done on Chrome and CocoonJS for Android.

obiot:
@parasyte i'm hesitating here, do we merge this as part of 2.1, or do we keep it for 2.2 ?
]

parasyte:
I haven't tested it, BUT I WANT IT.

obiot:
ahah, i want it too :P so then in 2.1 ?

I'll let you do the honor of clicking on the merge button as your confirmation you want it now :)

After that we should enter into feature freeze for 2.1, else we will never release it :P

parasyte:
@obiot @TheManuz merged!

I also fixed a few bugs along the way.

obiot:
awesome ! thanks ;)

TheManuz:
@parasyte thanks for doing the merge and fix bugs you encountered!
Sadly, i've got very little time lately, so i couldn't work on this! (I started the branch during christmas holidays, i've had much time then!).
Anyway, i'm happy! :smile:

obiot:
@TheManuz thank you again for this great contribution :)

Issue #616, opened 2014-12-06
Expect case-sensitive names


parasyte:
There are a lot of internal checks and transformations that occur to make names (even file names) lower case to support case-insensitive matching. This is bad for many reasons. The biggest reason is that JavaScript itself is a case-sensitive language. And second, we could save a few bytes and CPU cycles by removing all of the `toLowerCase` calls scattered through the engine.

I won't belabor the point; this ticket is for discussing whether we agree that case-insensitive matches are a bad idea, and if we can get rid of them for the next major release.

obiot:
Is this one done then?

agmcleod:
Don't think so. There are some places in TMX files i noticed that still use it. I wasn't quite sure about changing those, if that's just how tiled does it sort of thing. Other then that i believe it is done.

parasyte:
Fixed the last of these! There are only four occurrences of `to(Lower|Upper)Case` in the source tree, and the only one left that deals with string comparisons is checking for "collision" in TMX layer names.

PR #636, opened 2015-01-17
Added me.pool.exists & allowed for the class to be set in Tiled objects

xorinzor:


xorinzor:
Don't merge this yet, seems to have some issues. Working on it.

src/system/pooling.js
```diff
@@ -172,6 +172,18 @@
...
+        /**
+         * Check if an object with the provided name is registered
+         * @name exists
+         * @memberOf me.pool
+         * @public
+         * @function
+         * @param {String} name of the registered object
+         */
+        api.exists = function (name) {
+            return typeof(entityClass[name]) === "object";
```
parasyte:
``` javascript
return name in entityClass;
```

src/level/LevelDirector.js
```diff
@@ -111,7 +111,7 @@
                         );
                     } else {
                         obj = me.pool.pull(
-                            "me.Entity",
+                            me.pool.exists(settings.className) ? settings.className : "me.Entity",
```
parasyte: Is this check really necessary? Or can we just get away with:

``` javascript
settings.className || "me.Entity"
```

xorinzor: that could work too, I just took the safe route to prevent any possible exceptions

parasyte: And actually, it should use `settings.name` to match the non-collision-layer entities.

xorinzor: Isn't the collision name only for the layer-group? I've got several objects within the collision group with the name "solid" that work just fine

parasyte: Nah, I don't think the name is used any more since the collision layer update.

src/system/pooling.js
```diff
@@ -172,6 +172,18 @@
...
+        /**
+         * Check if an object with the provided name is registered
+         * @name exists
+         * @memberOf me.pool
+         * @public
+         * @function
+         * @param {String} name of the registered object
+         */
+        api.exists = function (name) {
+            return name.toLowerCase() in entityClass;
```
parasyte: We're no longer forcing lowercase for these kinds of lookups (see #616)

src/system/pooling.js
```diff
@@ -172,6 +172,18 @@
...
             entityClass[name].pool.push(obj);
         };
+        
+        /**
+         * Check if an object with the provided name is registered
+         * @name exists
+         * @memberOf me.pool
+         * @public
+         * @function
+         * @param {String} name of the registered object
```
parasyte, line 183: It actually returns a `Boolean` type.

xorinzor, line 183: for the jsdoc I just copied it from your other methods, I couldn't find a @return in any of the other methods so I left that out

https://github.com/xorinzor/melonJS/blob/master/src/system/pooling.js#L59

src/system/pooling.js
xorinzor: I was testing my code with 2.0.2, didn't notice that

obiot:
so is this one "ready" for merging, or still needs more love ?

xorinzor:
Pretty sure it's good to go

parasyte:
Yep, looks good to me.

PR #641, opened 2015-02-02
New fullscreen scaling mode

Djokal:
me.video.init no longer takes maintainAspectRatio as argument, it has been replaced with scaleMethod which accepts:

"default" EDIT: it's have been renamed to "fit"
The default setting where scale is fixed with black banners added if window scale is different. (the old maintainAspectRatio=true)

"stretch"
This fits the entire canvas on the screen without respecting canvas scale. (the old maintainAspectRatio=false)

"fill"
this will fill the entire screen with the canvas while respecting the scale, instead of adding black banners, a part of the canvas is not visible.

NOTE: I tested it and it works perfectly, but it still needs to resize the canvas. Might need some pointer on this


parasyte:
Please fix the build failure: https://travis-ci.org/melonjs/melonJS/builds/49241970 We have a really strict .jshint config file that enforces whitespace rules. :) Most of the warnings are very simple. The variable redefinitions can be fixed by moving the `var` statements outside of the if-blocks.

Djokal:
Done ;)

src/video/video.js
```diff
@@ -109,9 +109,9 @@
          * @param {Number} [options.renderer=me.video.CANVAS] renderer to use.
          * @param {Boolean} [options.doubleBuffering=false] enable/disable double buffering
          * @param {Number|String} [options.scale=1.0] enable scaling of the canvas ('auto' for automatic scaling)
-         * @param {Boolean} [options.maintainAspectRatio=true] maintainAspectRatio when scaling the display
+         * @param {Boolean} [options.scaleMethod="default"] ('default','fill','stretch') screen scaling modes
          * @param {Boolean} [options.transparent=false] whether to allow transparent pixels in the front buffer (screen)
-         * @param {Boolean} [options.antiAlias=false] wheter to enable or not video scaling interpolation
+         * @param {Boolean} [options.antiAlias=false] whether to enable or not video scaling interpolation
```
parasyte, line 114: Bonus typo fix!

src/video/video.js
```diff
@@ -136,7 +136,7 @@
             // sanitize potential given parameters
             settings.doubleBuffering = !!(settings.doubleBuffering);
             settings.autoScale = (settings.scale === "auto") || false;
-            settings.maintainAspectRatio = !!(settings.maintainAspectRatio);
+            settings.scaleMethod = !!(settings.scaleMethod);
```
parasyte: This transforms the value into a boolean. I don't think you want to do that with a string input.

Kaldjo: Shall I changed to?
settings.scaleMethod = (settings.scaleMethod === "fill" || settings.scaleMethod === "stretch") ? settings.scaleMethod : "default";

parasyte: It'll work, but you're already doing an if-then-else check later in the code, and never _explicitly_ checking for "default" anyway.

You can also use a little trick with an array:

``` javascript
settings.scaleMethod = [ "fill", "stretch" ].indexOf(settings.scaleMethod) >= 0 ? settings.scaleMethod : "default";
```

Kaldjo: Nice :D.

parasyte:
@Djokal I really like the sounds of this! :smiley: I haven't had a chance to try it yet, but everything else looks good so far.

src/video/video.js
```diff
@@ -340,23 +340,34 @@
...
                 var _max_height = Math.min(maxHeight, parent.height || window.innerHeight);
-
-                if (settings.maintainAspectRatio) {
-                    // make sure we maintain the original aspect ratio
-                    var designRatio = me.video.renderer.getWidth() / me.video.renderer.getHeight();
-                    var screenRatio = _max_width / _max_height;
+                var designRatio = me.video.renderer.getWidth() / me.video.renderer.getHeight();
+                var screenRatio = _max_width / _max_height;
+                var sWidth = me.video.renderer.getHeight() * screenRatio;
+                var sHeight = me.video.renderer.getWidth() * (window.innerHeight / window.innerWidth);
```
obiot: why use `window.innerHeight` and `window.innerWidth` here instead of the previous `_max_width` and `max_height` ? the former provides the max size of the window, the later provide the max size of the document element in which the canvas is, or if none the window size. 

note that I'm not sure my self, just asking to see if you are clearer on your side about it :P

obiot:
love it, super useful pull request !

obiot:
as a side note, this is no equivalent to the avaialble cocoonJS options :
http://support.ludei.com/hc/en-us/articles/202515383-Scale-modes-in-canvas-

* `absolute` :  scaling off
* `scale aspect fit` : scaling on & scaleMethod `default`
* `scale aspect fill` : scaling on & scaleMethod `fill`
* `scale fill` : scaling on & scaleMethod `stretch`

src/video/video.js
Kaldjo: I made this pull request because I wanted to make a game fullscreen on mobiles. I wasn't sure about the other ones.
shall I change it?

obiot: yes I would to be honest, especially since you already use `screemRatio` for `sWidth` that already takes them in account  (only sHeight then is not consistent) : 

``` javascript
var screenRatio = _max_width / _max_height;
```

then for me, good to merge as well ;)

Djokal:
@obiot The link is not working for me, can you please elaborate?

parasyte:
It's an issue with Ludei's ZenDesk at the moment.

obiot:
the link just list the supported scaling mode by cocoonJS, and i was just saying that with your pull request, we have now something equivalent :)

src/video/video.js
Kaldjo: I'll add it right away, I am working on an example I'll be able to host it in couple minutes.
I still have to find out how to resize the canvas to the screen before merging it.

Djokal:
Good to hear :), 
Here's a quick and dirty example.
https://github.com/Djokal/Melonjs-FillScreen-Example

parasyte:
Looks like the Ludei link is working again.

obiot:
uh !
````javascript
(!me.video.init("screen",  me.video.CANVAS, 960, 640, true, 'auto',"fill"))
````

did not try your example but it seems to stil use the old video.init function signature :P

**Comment by Djokal** [comment]
It's the new one, I kept the scale just in case anyone wants to set it manually.

api.init = function (wrapperid, renderer, game_width, game_height, doublebuffering, scale, scaleM)

**Comment by obiot** [comment]
no it's not, or I"m totally crazy (also possible),

With your pull request we should now have :
````javascript
// init the video with a 640x480 canvas
me.video.init(640, 480, {
    wrapper: "screen", 
    renderer: me.video.CANVAS,
    scale: 'auto', 
    scaleMethod: "fit",  <<< this is what you changed
    transparent: true
});
````

**Comment by Djokal** [comment]
Be assured, you're not crazy I am :laughing: . I commited using the old 2.0.2 version. I'll update it in a minute.

**Comment by Djokal** [comment]
It's up to date now.

**Comment by Djokal** [comment]
after thinking about it, It's the vertical scrollbar that's causing the problem by reducing the window width !!!

**Comment by Djokal** [comment]
Quick and dirty temporary fix.
overflow is hidden, but canvas not resized yet, I'll look into it tomorrow.

**Comment by Djokal** [comment]
Quick and dirty temporary fix.
overflow is hidden, but canvas not resized yet.

here is a hosted example: http://melonjs-fillscreen-example.paperplane.io/

**Comment by Djokal** [comment]
@parasyte @obiot I think it's ready now.
Online example : http://melonjs-fillscreen-example.paperplane.io/
Example repository  : https://github.com/Djokal/Melonjs-FillScreen-Example

**Comment by parasyte** [comment]
@Djokal Do you mean "static width" and "static height"? I think those would be very useful. The current code treats both width and height dynamic, depending on the orientation:

* Landscape orientations use a static width
* Portrait orientations use a static height

In both cases, the dynamic dimension is only allowed to shrink (it can never grow beyond the size passed to `me.video.init`). What I would propose with the "static width"/"static height" modes is that the dynamic dimension is always allowed to shrink or grow. As an example, the "static height" mode would be most useful for side-scrollers, where more of the map is revealed on wider screens. Similarly, "static width" is useful for vertical shooters, where more of the map is revealed on taller screens.

**Comment by Djokal** [comment]
@parasyte good idea, I'll try to add those modes.

**Comment by obiot** [comment]
One more thing, what about renaming "default" to "fit" just to be more explicit about the default behavior ?

**Comment by Djokal** [comment]
@obiot me.video.renderer.getWidth() is just returning the width, I need the equivalent of canvas in webGL

**Comment by parasyte** [comment]
@Djokal WebGLRenderer has a `getCanvas` method. You should be careful about changing its size, though, since there's also the `dimensions` property and Compositor projection matrix that need to stay in sync. See `me.WebGLRenderer.resize()`

In fact, maybe that method needs to be renamed to `scale` and a new `resize` method created to do all of the canvas resizing work with pixel dimensions.

**Comment by obiot** [comment]
@Djokal both renderer have a `getWidth()` and `getHeight()` function inherited from the base renderer class, that's what I was trying to explain. 

https://github.com/melonjs/melonJS/blob/master/src/video/renderer.js#L202
https://github.com/melonjs/melonJS/blob/master/src/video/renderer.js#L213

so there is no need to use the `getCanvas()` function to get the unscaled size.

**Comment by Djokal** [comment]
@parasyte Indeed I was thinking about the resize function too, am a bit overwhelmed with all the informations, just give me some time to digest all of this haha.

@obiot good point, but I am trying to set the height and width of the canvas here.
    var backbuffer = me.video.renderer.getCanvas();
    backbuffer.width = _max_width / scaleX;
    backbuffer.height = me.video.renderer.getHeight();

**Comment by Djokal** [comment]
@parasyte About renaming resize to scale, webglRenderer already has a scale function
        scale : function (x, y) {
            this.globalMatrix.scale(x, y);
        },
Shall I rename that one too scaleGlobalMatrix?

**Comment by parasyte** [comment]
@Djokal Ah, you are correct! The CanvasRenderer also has a `scale` method. In this case, the `resize` should probably be renamed `scaleCanvas` or such. And the new method named `resizeCanvas`. Does that seem reasonable?

**Comment by Djokal** [comment]
Yes, that's better.

File: src/video/video.js [hunk 4]
```diff
@@ -70,7 +70,7 @@
         });
 
         /**
-         * Select the HTML5 Canvas renderer
+         * Select the HTML5 Canvas rendererg
```
parasyte: Remove this change, please

src/video/canvas/canvas_renderer.js
```diff
@@ -253,6 +253,19 @@
...
+        /**
+         * resizes the canvas
+         * @name resize
+         * @memberOf me.CanvasRenderer
+         * @function
+         */
+        resize : function (width, height)
+        {
+            this.backBufferCanvas.width = width;
+            this.backBufferCanvas.height = height;
```
parasyte: This can be moved to `me.Renderer` if you use:

``` javascript
var canvas = this.getCanvas();
canvas.width = width;
canvas.height = height;
```

Then remove the WebGLRenderer.resize method (both classes extend `me.Renderer`)

src/video/video.js
```diff
@@ -109,9 +109,9 @@
          * @param {Number} [options.renderer=me.video.CANVAS] renderer to use.
          * @param {Boolean} [options.doubleBuffering=false] enable/disable double buffering
          * @param {Number|String} [options.scale=1.0] enable scaling of the canvas ('auto' for automatic scaling)
-         * @param {Boolean} [options.maintainAspectRatio=true] maintainAspectRatio when scaling the display
+         * @param {Boolean} [options.scaleMethod="default"] ('default','fill','stretch') screen scaling modes
```
parasyte: Documentation: "default" needs to be renamed "fit"

src/video/video.js
```diff
@@ -119,7 +119,7 @@
          *       wrapper: "screen",
          *       renderer: me.video.CANVAS,
          *       scale: 'auto',
-         *       maintainAspectRatio: true,
+         *       scaleMethod: 'default',
```
parasyte: Documentation: "default" needs to be renamed "fit"

obiot:
shall we merge #641 into master ? what's left to be done on that one ?

also as I was proposing last time, shall we rename `grow-width` and `grow-height` to `resize-width` and `resize-height` or event `resize-horizontal` and `resize-vertical` to match the similar CSS properties ?

parasyte:
This is now merged. Thanks again @Djokal If you spot any regression on master, please create a new ticket or pull request.

obiot:
We already closed this one, but based on the following i'm wondering if we should not go a bit further in terms of renaming when it comes to all the fit/fill scale modes :

https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit

parasyte:
So you suggest something like:

* `scaleMethod` -> `screenFit`
* `fit` -> `contain`
* `flex-width` -> `cover-height`
* `flex-height` -> `cover-width`

The `flex` modes need to be oppositely named, because "flex" means the dimension is dynamic, and "cover" means the dimension is constant. E.g. `"dynamic-width" == "constant-height"`

I kind of like the shorter names because they are shorter. :) But if it will help documentation and user familiarity, it's a good name change.

obiot:
I would actually keep flex as these two involves resizing the viewport, which is not the case i think for thr fit(s) mode.

Then i would do :
- fit -> fit-contain
- fill-max -> fit-fill
- stretch -> fit-scale-down

I'm wondering as well if it's make sense to keep the "scale" setings, and if we could not directly use that setting instead of an additonal "scaleMode" one.

parasyte:
@obiot Yeah sorry, I updated the comment above.

I don't think we want to prefix all of the values with "fit". That's not the way it's done in the CSS.

obiot:
Indeed, but i thought would fit well as way to differentiate the "flex" modes and the "fit"' modes.

obiot:
so to finish with a clean suggestion on this one:
- `fit-contain` (former `fit`)
- `fit-fill` (former `fill-max`)
- `fit-???` (former `fill-min`)
- `fit-cover` (former `fill-strech`, but not sure about this one?)
- `flex-width` (no changes)
- `flex-height` (no changes)
- `flex-both` (to be added)

On top of this (and that's why I'm suggesting to add/use either the `fit-` or `flex-` prefix) , i would also remove the current `scale` property, so that instead of having to specify `auto` for scale and then add the scaling mode through `scaleMode`, we can just use scale that can contains either a numeric value, or one of these strings.

that also brings another changes, which i think makes sense, as if scale is not specified we then use by default the `fit-contain` mode, as opposed to no scaling at all righ now.

parasyte:
I can't agree with combining the scale (numeric) and method, because the `flex-both` mode will work with scaling factors other than 1.0 The method does not scale, so a scaling factor is optional.

Also "fit" is not an ideal prefix. Since these are all scaling modes, you could opt to prefix them with "scale". I'm still not a fan of prefixing these, though.

obiot:
okay, no changes then :)

Note that if we do have the `flex-both` mode in the future, we can just call it `flex` as well

PR #647, opened 2015-02-09
updated shobox settings to work with latest shoebox

terebentina:


parasyte:
Nice addition! Thank you.

obiot:
hi, just to be clear since there is a format change, what minimum version is required for this one ? do you know, or shall we just say 3.5.4 minimum ?

obiot:
note, I also updated the related wiki page :
https://github.com/melonjs/melonJS/wiki/ShoeBox

terebentina:
No idea when it changed, unfortunately. I upgraded shoebox before trying
the settings and then when I tried it said something like "import partial"
with your original setting.
Btw: I can make changes to your wiki pages - is this intended?

obiot:
oh ok, not a big deal anyway, just update to the latest one is still good information :)

for the wiki, i already updated it, but thank you !

Issue #649, opened 2015-02-13
Investigate translation bug with nested floating renderables


parasyte:
This issue pops up all the time on the forum/in chat. A floating Renderable inside a floating Container will move with the viewport in the opposite direction (it's weird). This is known to happen in 2.0.x. Ticket is to track its status for 2.1

obiot:
does this still happen with all the changes made by Aaron ?

agmcleod:
Yes. It doesn't handle the floating state very well.

parasyte:
I forgot to link the forum post in question: https://groups.google.com/d/msg/melonjs/AFqhE1u3G2U/g9cnV7eTaRsJ

agmcleod:
2504f2e

Added a version of the updateAbsoluteBounds() to the container that invokes the resize method. It still uses a child bounds object, as I figure it's better than constructing a new Rect each time :). No longer runs on add/remove child calls.

Not sure what you mean by clipping regions in the commit, but this is a bit of a simpler version.

parasyte:
Canvas clipping regions:

```javascript
var context = renderer.getContext();
context.save();
context.clip(this.pos.x, this.pos.y, this.width, this.height);
// ... loop through children here and call their draw method
context.restore();
```

https://developer.mozilla.org/samples/canvas-tutorial/6_2_canvas_clipping.html

For WebGL, use the stencil buffer.

agmcleod:
Wouldn't this have the same problem of drawing objects outside the container dimensions? Or if we're not worried about that, can even just keep the inViewport test against the container itself, and not worry about updating the rect.

aaschmitz:
@agmcleod with the latest adjustments, the particles examples show the error below in console:

TypeError: this.getBounds(...) is undefined melonJS-2.1.0.js:8213

agmcleod:
Yeah i didn't do any changes to that one yet. It's probably still broken in webgl for the regular debugger too. Was just making some fixes.

parasyte:
@agmcleod My argument is that it would simplify the melonJS code to clip all children that are outside of their container bounds, instead of attempting to iterate all children recursively to compute the minimum bounding box and resize it on every frame.

obiot:
same here, except someone is super motivated to fix within the next couple of days ?

parasyte:
IMHO, this should be moved to the next major milestone (3.0.0). Opinions?

obiot:
[...]

PR #657, opened 2015-02-20
fix for tile objects with empty name

terebentina:


terebentina:
From chat with @parasyte :
parasyte an hour ago 
"TileObject" is set here: https://github.com/melonjs/melonJS/blob/v2.0.2/src/level/TMXObjectGroup.js#L333
And defined in the pool here: https://github.com/melonjs/melonJS/blob/v2.0.2/src/system/pooling.js#L43
This is what I meant by "hardcoded"
Tile Object with no name -> creates a Sprite.
terebentina 16 minutes ago 
aha!!!!
had to debug melon step by step but I found the problem
TileObject is indeed set here https://github.com/melonjs/melonJS/blob/v2.0.2/src/level/TMXObjectGroup.js#L333
IF the name is undefined
however, it's not undefined, it's the empty string
at least in tiled 0.11

parasyte:
@obiot @agmcleod We might actually want do a 2.0.3 release? This patch and maybe one for the floating container bug #649 would be a nice for the 2.0.x folks, even with the next release coming soon...

What do you think?

obiot:
Hmmm... I would say: 
Is there any easy workaround to those two issues? If yes i would not do it as we are on the verge anyway to release 2.1.0

obiot:
Not sure though that the fix is correct on 2.0.2, as this would probably set the name of the collision shapes also to "TileObject" (would work though)

parasyte:
@obiot With all the bugs we are fixing (and still finding) I don't think 2.1 is nearly ready for RC.

obiot:
indeed, we have more & more issues popping up :(

obiot:
2.1.0 being there now, closing this one :P

PR #658, opened 2015-02-20
fix: mouse-wheel events were ignored

ldd:


src/input/pointer.js
```diff
@@ -354,6 +354,17 @@
...
+                    case MOUSE_WHEEL:
+                        // event inside of bounds: trigger the MOUSE_WHEEL callback
+                        if (eventInBounds) {
+                            // trigger the corresponding callback
+                            if (triggerEvent(handlers, e.type, e, e.pointerId)) {
+                                handled = true;
+                                break;
+                            }
+                        }
+                        break;
```
parasyte, line 356: This looks identical to `POINTER_DOWN`. You should be able to add `case MOUSE_WHEEL:` above for a smaller patch.

parasyte, line 356: Or use `default:` here and delete the `POINTER_DOWN` case. I kind of prefer that, TBH. It may be more future-proof.

ldd:
done. If I recall correctly, the original code was also future-proof

parasyte:
Perfect! Thanks for debugging and patching. And yeah, the original code didn't use a switch statement for event dispatching. :smile:

Issue #690, opened 2015-06-03
Followup: Better release process


parasyte:
Followup to #508 

We should consider the following options:

* [ ] Add a new grunt task to bump the version
* [ ] Release non-versioned build files

The former is pretty obvious, and could be done as part of `grunt release`.

The latter is more of a political matter. All popular frameworks release build files without a version number in the filename. This allows easy updates for minor versions; `bower update` is all you need. Or even CDN releases that automatically propagate to CDN users. (But this is dangerous, as any changes may introduce new bugs that break CDN games without warning!)

PR #693, opened 2015-06-10
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,5 @@
 {
   "name": "melonJS",
-  "version": "2.2.0",
   "main": "build/melonJS-2.2.0-min.js",
   "repository": {
     "type": "git",


parasyte:
@kkirsche I'm going to close this one. The version property will be used in the future. We are instead going to remove the version number from the file name in #690

kkirsche:
From the discussions I've had it isn't going to be used and most recommendations are to use NPM. The original reason for this was for a package management site. This never happened and no one cared to make it. Personally, I think it's a waste as most discussions re future dev have been around it's slow and painful death. 

Best of luck though

parasyte:
I'm unfamiliar with the backstory. Until it's removed from the bower docs, it doesn't hurt to keep it.

kkirsche:
Also their maintainer says they probably won't ever use it: http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

parasyte:
Alright, I'll remove it. Thank you for the patch.

PR #695, opened 2015-06-12
Fix for flat-top hexagonal rendering bug

matthewmmorrow:
Corrected staggering logic for flat top hexagons. Will add a flat-top example later to show that the rendering works fine for both types of rotation but wanted to get this bug fix in ASAP.


obiot:
+1 :P

PR #696, opened 2015-06-12
Fix tiled object rendering position in hex

matthewmmorrow:
Double checked all hex renderer methods are implemented.

### DIFF
--- a/src/level/TMXRenderer.js
+++ b/src/level/TMXRenderer.js
@@ -525,8 +525,13 @@
          * the way Tiled places them
          * @ignore
          */
-        adjustPosition: function (/*obj*/) {
-            // TBD
+        adjustPosition: function (obj) {
+            // only adjust position if obj.gid is defined
+            if (typeof(obj.gid) === "number") {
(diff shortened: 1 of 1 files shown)


src/level/TMXRenderer.js
```diff
@@ -525,8 +525,13 @@
          * the way Tiled places them
          * @ignore
          */
-        adjustPosition: function (/*obj*/) {
-            // TBD
+        adjustPosition: function (obj) {
+            // only adjust position if obj.gid is defined
+            if (typeof(obj.gid) === "number") {
```
agmcleod, line 530: @matthewmmorrow what's the reasoning for an object requiring gid?

obiot, line 530: there is some weird things about tiled object that have a wrong position in orthogonal maps, I suppose it is the same with hexagonal maps ?

@matthewmmorrow did u try yet to add some object to your map in Tiled and see if they were placed properly in melonJS ?

obiot:
Merging this one, assuming it's also needed to fix object positioning.

PR #702, opened 2015-07-07
Patched Object.extend simple inheritance to fix conflict with Facebook SDK

### LINKED ISSUE #666: Incompatible with underscore/lodash
I discovered that melonJS breaks underscore/lodash, because a module I wanted to use has such a dependency. The issue is an incompatible implementation of `Object.prototype.extend`.

Here is a workaround for node.js, but it is not applicable to browser environments... where `pathwatcher` is the module with the underscore dependency.

```javascript
var vm = require("vm");
var pathwatcher = vm.runInNewContext('require("pathwatcher")', { "require" : require });
```

### LINKED ISSUE #699: Conflicts with Facebook SDK
Some of the javascript monkey patching in melonJS conflicts with the Facebook SDK because they're doing some similar things in the SDK, and causes sporadic loading behavior.  The SDK will not load at all in Firefox and sometimes loads in Chrome/Safari, and sometimes not.

Specifically, when I comment out lines 219 to 297 of the following snippet, the Facebook SDK loads perfectly again: https://github.com/melonjs/melonJS/blob/970927095d77c4672f181a05d3d33b79454829b0/src/lang/object.js

Of course, then melonJS no longer works!  Is this an issue you all have run into before?  Any ideas on a workaround for this?  Thanks in advance!

jdrorrer:
Issue: Object.extend in melonJS conflicts with the Facebook SDK causing the SDK to not load properly. melonJS games that try to use the Facebook SDK with melonJS are not able to do so.

Fixes #699. Related to #666. Changes include...

- attached extend property to me.Object rather than Object (i.e. me.Object.extend)
- attached me.Object.extend to Error.extend
- changed all instances of Object.extend to me.Object.extend

Also...
- I updated this jsFiddle with the new changes as well: http://jsfiddle.net/jdrorrer/7z3t2o4s/
- all tests are passing

screen shot 2015-07-07 at 11 49 13 am

The issue can be recreated by forking and building a game built on melonJS that uses the Facebook SDK. A good example of this is Clumsy Bird: https://github.com/ellisonleao/clumsy-bird. It has a Facebook share button when the game ends.

Side note (maybe for a future update):
However, I think an ideal fix (one that should fix #666 completely) would be similar to how string.js uses `S()`: http://stringjs.com/.  melonJS could use the me global and attach all javascript language extensions to it, so that you could call extend like this: `me(Rect).extend`. string.js also has an `extendPrototype()` method. Something like that in melonJS would allow developers to add one line of code in their game and still use `Object.extend()`, `me.Rect.extend()`, etc. the same as they use it today.

### DIFF
--- a/src/lang/window.js
+++ b/src/lang/window.js
@@ -22,6 +22,19 @@
      */
     window.me = window.me || {};
 
+    /**
+    * Empty BaseClass constructor and prototype attached to 
+    * `me` global object to act as the base class for inheritance 
+    */
+    me.BaseClass = function () {};
+    me.BaseClass.prototype = Object.create(Object);
+
+    /**
+    * Instantiate new me.BaseClass and attach to me.Object
+    * to allow new sub-classes to be created using me.Object.extend
+    */
+    me.Object = new me.BaseClass();
+
     /*
      * DOM loading stuff
      */
--- a/src/lang/window.js
+++ b/src/lang/window.js
@@ -23,17 +23,14 @@
     window.me = window.me || {};
 
     /**
-    * Empty BaseClass constructor and prototype attached to 
+    * me.Object constructor and prototype inherits from Object and is attached to 
     * `me` global object to act as the base class for inheritance 
     */
-    me.BaseClass = function () {};
-    me.BaseClass.prototype = Object.create(Object);
-
-    /**
-    * Instantiate new me.BaseClass and attach to me.Object
-    * to allow new sub-classes to be created using me.Object.extend
-    */
-    me.Object = new me.BaseClass();
+    me.Object = function () {
+        Object.apply(this, arguments);
+    };
+    me.Object.prototype = Object.create(Object.prototype);
+    me.Object.prototype.constructor = me.Object;
 
     /*
      * DOM loading stuff
--- a/examples/particles/js/editor/controls.js
+++ b/examples/particles/js/editor/controls.js
@@ -1,7 +1,7 @@
 (function() {
     var pe = game.ParticleEditor = game.ParticleEditor || {};
 
-    pe.EmitterList = Object.extend({
+    pe.EmitterList = me.Object.extend({
         init : function(emitterController, container) {
             this.emitterController = emitterController;
             this.emitters = [];
@@ -137,7 +137,7 @@
         }
     });
 
-    pe.EmitterController = Object.extend({
+    pe.EmitterController = me.Object.extend({
         init : function(container) {
             this.widgets = [];
             this.ingameWidgets = [];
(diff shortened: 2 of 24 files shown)


src/lang/window.js
```diff
@@ -22,6 +22,19 @@
      */
     window.me = window.me || {};
 
+    /**
+    * Empty BaseClass constructor and prototype attached to 
+    * `me` global object to act as the base class for inheritance 
+    */
+    me.BaseClass = function () {};
+    me.BaseClass.prototype = Object.create(Object);
```
parasyte: Why do we need this empty class instead of just calling it `me.Object`?

parasyte:
Thanks @jdrorrer! It's exactly what I wanted for #666

Turning `me` into a callable for JS language extensions should be fine. There are a few global objects whose prototypes are given custom properties; `Array`, `Number`, and `String` for example.

src/lang/window.js
jdrorrer: To me, it made sense to separate the two. It's really not necessary, though. These were my thoughts...
1. Define the `extend` property on a `BaseClass` for simple 'class' inheritance - so all new classes inherit from `me.BaseClass.prototype`.  
2. Assign a `new me.BaseClass()` to `me.Object` so that `me.Object.extend()` could be used in place of `Object.extend()`.  Doesn't have to be assigned to `me.Object` either, but I just thought that would be the easiest transition from `Object.extend()`.

jdrorrer:
@parasyte you're welcome, glad I could help!

Okay, I see... I was thinking some of those custom properties may be interfering with underscore/lodash as well, but maybe not?

src/lang/window.js
parasyte: `me.Object` is exactly what I had in mind, no changes there. If there's no technical reason that `me.Object` can't be the base class, I would prefer that.

parasyte:
•shrug• It's a possibility that they conflict with other libraries, including underscore and lodash. I don't have evidence either way. Just listing some potential issues for followup patches, since I agree with your suggestion.

src/lang/window.js
jdrorrer: Nope, no technical reason and that should be an easy fix!

jdrorrer:
Okay, I see again... lol.  Very cool, thanks man!

src/lang/window.js
jdrorrer: Okay, the update is ready to go!  It seems like it performs slightly better, too for some reason...

parasyte:
Perfect! I'll push some followup commits for CHANGELOG. Thank you for the contribution!

obiot:
Awesome, that's also what I had in mind to address these compatibility issue, super happy to see someone stepped in and got it done :)

Great contribution, thank you very much !

jdrorrer:
Ahh shucks, thanks guys! Glad to help! :)

Issue #390, opened 2014-01-23
Better animation definitions


parasyte:
Animations implicitly use the same same "speed" property for every frame of the animation. Even the simplest animated gifs have the ability to define time delays on each individual frame. This provides some additional control in how an animation plays.

For example, you might be animating a character's attack, which needs a short "pause" cycle to telegraph the coming attack (which a skilled player will identify, and be able to respond with a block or counter-attack). The pause cycle may actually be two or three frames long, while the rest of the animation plays at full speed. With today's animation definitions, an animation like this must be defined as:

```javascript
{
    "frames" : [ "telegraph", "telegraph", "telegraph", "punch1", "punch2", "punch3" ],
    "speed" : 30
}
```

Where the `"telegraph"` frame is duplicated to create the illusion of a short (~90ms) pause. We can make this better by setting the delay on a per-frame basis:

```javascript
{
    "frames" : [
        { "name" : "telegraph", "delay" : 100 },
        { "name" : "punch1", "delay" : 30 },
        { "name" : "punch2", "delay" : 30 },
        { "name" : "punch3", "delay" : 30 }
    ]
}
```

This example shows that the `"telegraph"` frame will be displayed for at least 100ms, and the other three frames for at least 30ms each.

Also notice that the single frame names have been replaced with a dictionary; we can include extra per-frame meta data here, like scale, angle, alpha, background color, composite operation, ... Just for much better control over how each frame is drawn. Some of these can also be Tweened :warning: for smooth animations.

parasyte:
This seems like a good place to mention a minor bug reported on the forum. The animation frame timing is a "minimum frame timing", and needs to be turned into an "absolute frame timing" by taking into account the total elapsed time, and picking the correct frame to display. Right now the next frame in the sequence is always chosen, even if the frame rate would have caused animation frames to be dropped.

See: https://groups.google.com/d/msg/melonjs/mM4BdyQyNY8/ygEJUIPvDJwJ

obiot:
indeed that makes sense, although if ones lower the frame rate, and if the animation speed is not adjusted it would result in skipping frame, might be confusing/disturbing as well I suppose ?

obiot:
so this one is done ?

obiot:
i guess we can properly fix the tmx tile animation stuff then now :
https://github.com/melonjs/melonJS/blob/master/src/level/TMXTile.js#L138

:)

parasyte:
No, this one's not nearly done.

[...]

PR #707, opened 2015-07-12
Ticket 390

agmcleod:


src/renderable/animationsheet.js
```diff
@@ -143,7 +143,7 @@
                             "string parameters for addAnimation are not allowed for standard spritesheet based Texture"
                         );
                     } else {
-                        this.anim[name].frame[i] = this.textureAtlas[this.atlasIndices[index[i]]];
+                        this.anim[name].frame[i] = Object.assign(this.textureAtlas[this.atlasIndices[index[i]]], { delay: this.anim[name].animationspeed });
```
parasyte, line 146: I think this should be:

``` javascript
this.anim[name].frame[i] = Object.assign(
    {},
    { "delay" : this.anim[name].animationspeed },
    this.textureAtlas[this.atlasIndices[index[i]]]
);
```

Based on the properties of `Object.assign`;
- Don't update the underlying atlas structure (first arg is empty Object)
- Don't override `delay` property if it exists on the atlas (`{ "delay" : ... }` Object before frame Object)

And also broken into multiple lines, because it's getting kind of long.

(Should have the same change on the diff above.)

agmcleod, line 146: First point,  i didn't implement it this way, as I find it odd that we would have the texture atlas object contain delay information, when that is purely about image coordinates & rotation.

Second point, agreed it should be split up on a couple lines :)

parasyte, line 146: :+1:

agmcleod:
Tagging the issue #390

Issue #715, opened 2015-07-29
Create an origin point in each frame of a texture atlas that determines the frame's position relative to the coordinates it's drawn at.


Giwayume:
Basically the rundown is:
1. The "anchor point" becomes a point that scaled from the entity's x and y position (0,0) to the entity's width and height (1,1). Similar to today.
2. Each animation (atlas) frame gets a new property, the origin point (a pixel offset from the "offset" point), that is defined by the pivot point in TexturePacker (or shoebox equivalent, or manually, whatever). The origin point is drawn at the entity's anchor point, with the rest of the texture drawn relative to that origin point.
3. Scaling and rotation transformations happen around each frame's origin point.

This change can impact existing projects. So maybe an opt-in? 3.0?

Link to the original discussion:
https://groups.google.com/forum/embed/?place=forum%2Fmelonjs&showsearch=true&showpopout=true&hideforumtitle=true&parenturl=http%3A%2F%2Fmelonjs.org%2Fforum#!topic/melonjs/KYes1bzTsBU

Giwayume:
If the melonjs lords Oliver, Jason, Aaron agree with this direction, assign this issue to me?

agmcleod:
Can't assign to non collabs :). I agree with the issue. AnimationSheet & TextureAtlas should be the only classes changing I think. Be sure to run `grunt` after making any changes, fix any issues jshint or the tests complain about. Once you have a working scenario, and you have tested it against the platformer example (didn't break existing behaviour), make a pull request.

Giwayume:
How do I collab?

Also, by design it should break the platformer example because the pivot point exported from TexturePacker is at [0.5,0.5] but the platformer example sets the anchor point at [0.5,1]. Half the sprite should end up clipping through the floor without modifying the TexturePacker file.

agmcleod:
@Giwayume even if the anchorPoint needs to be changed, that's fine. That's a good learning in what other users might need to do in an upgrade guide.

For collaborating, fork the project. Push commits to your own repo. Do a pull request to our master.

Giwayume:
I'm seeing the "Texture" class inside of video/canvas/texture.js, what do you mean by the "TextureAtlas" class? Am I looking at old code?

agmcleod:
@Giwayume yes that's it. Sorry i had actually forgotten about the rename.

Giwayume:
OK, before I make a mess, I've noticed something.

[...]

Issue #309, opened 2013-11-18
me.Sprite.anchorPoint should specify position anchoring


parasyte:
Right now, sprites are always drawn with their upper-left corner, regardless of the `anchorPoint`, which is only used for scaling and rotation. `anchorPoint` should also be used for positioning.

obiot:
tagging this one for 3.0.0, following the changes on #715 !

parasyte:
ooooh, really?! This is excellent! I'll have to check out #715 tomorrow. :)

PR #716, opened 2015-07-30
[#715] texture atlas origins (anchorPoints)

Giwayume:
I created two new properties for a texture's atlas:

{boolean} hasTextureAnchorPoint - Whether or not the current frame defines its own anchor point.
{me.Vector2d} anchorPoint - The current frame's anchor point (0,0) top-left to (1,1) bottom-right.

If you're creating a custom AnimationSheet, you can pass an anchorPoint in addition to the framewidth and frameheight.

var animationSheet = new me.AnimationSheet(0, 0, {
    image : "animationsheet",
    framewidth : 64,
    frameheight : 64,
    anchorPoint : new me.Vector2d(0.5, 0.5)
});

If you're using TexturePacker, frame anchorPoints are created automatically based on the "sourceSize", "spriteSourceSize", and "pivot" properties of the frame (assuming they all exist).

If an anchorPoint is set by the texture atlas json or manually like above, the entity.renderable's anchorPoint is drawn at the entity.body's anchorPoint. (as described in issue 715)

Otherwise, it acts the same as it does today, where the entity attempts to keep the renderable inside the body's bounds.


src/video/canvas/texture.js
```diff
@@ -140,12 +140,22 @@
...
                     atlas[frame.filename] = {
-                        name    : name, // frame name
-                        offset  : new me.Vector2d(s.x, s.y),
-                        width   : s.w,
-                        height  : s.h,
-                        angle   : (frame.rotated === true) ? nhPI : 0
+                        name         : name, // frame name
+                        offset       : new me.Vector2d(s.x, s.y),
+                        hasTextureAnchorPoint : hasTextureAnchorPoint,
+                        anchorPoint  : (hasTextureAnchorPoint) ? new me.Vector2d(originX / s.w, originY / s.h) : null,
```
obiot, line 154: we don't really need a `hasTextureAnchorPoint` boolean, we can just check if `anchorPoint` is defined or not ?

obiot:
Hi @Giwayume great submission! let me have a quick review and see if I ahve any questions on my side :P

src/entity/entity.js
```diff
@@ -276,14 +276,27 @@
...
+                else {
+                    // translate the renderable position (relative to the entity)
+                    // and keeps it in the entity defined bounds
+                    x = ~~(0.5 + this.pos.x + this.body.pos.x + (
+                        this.anchorPoint.x * (this.body.width - this.renderable.width)
+                    ));
+                    y = ~~(0.5 + this.pos.y + this.body.pos.y + (
+                        this.anchorPoint.y * (this.body.height - this.renderable.height)
+                    ));
+                }
```
obiot: wonder if there is any way to optimize this and get rid of the if/else statement. maybe @parasyte wil have some idea about it. Furthermore we had this #309 ticket, that maybe could be also integrated somehow through this changes you did?

src/video/canvas/texture.js
Giwayume, line 154: Yes, but there was a reason I put that in there... hmm let me remember.

src/entity/entity.js
agmcleod: Could dry it up by setting values to the return of the bits on lines 284, 287, 294, 297. Then do the repetitive `x = ~~(0.5 + this.pos.x + this.body.x + xAnchor)`. that bit could even be in a method. Dynamically pass the property you want. x or y.

src/video/canvas/texture.js
Giwayume: The problem that came up in my head was where one frame in an animation had an anchorPoint defined but another did not. But looking back having a separate boolean didn't exactly fix that problem. I'm not setting the original anchorPoint back to default again, so that problem isn't entirely fixed (rare case, though)

Giwayume: I think sticking in the boolean made it more externally obvious that this variable could possibly be null. I'll take it out, though.

obiot, line 154: note that i'm not saying that we should remove it, just wondering if it's really necessary, as the less properties we have the better, but in the mean time it's not like this one is anyway impacting performances of the object.

Giwayume, line 154: OH! I see. I'm not crazy. The AnimationSheet inherits an anchorPoint from the Renderable class. The boolean is there so we know whether the anchorPoint came from the texture, or whether it was the default anchorPoint. If it's from the texture, the sprite's position is drawn differently than it is today (reference entity.js)

src/entity/entity.js
Giwayume: I think #309 is covered here, however the positioning is handled inside of the Entity instead of the Sprite class.

src/video/canvas/texture.js
Giwayume, line 154: The boolean doesn't have to exist inside the atlas, but it does need to be defined when we get to the setAnimationFrame() function inside the AnimationSheet class.

src/entity/entity.js
obiot: it seems to be yes (although `me.Sprite` should be updated as well to then comply with the new behavior here), but to be honest I did not have the time yet to go through all your changes :P

Giwayume: If the anchorPoint translation moves to `me.Sprite`, it would be difficult for `me.Entity` to keep its current positioning behavior, where it completely ignores the `me.Sprite` anchorPoint and tries to position the edges of the sprite based on body width/height. 

I wasn't sure if it was OK to toss out the old positioning behavior. Some people may find it useful. Though, I think if we have both it makes the engine a little more confusing to use rather than committing to the new method. And having the Sprite positioned relative to Entity's body differently based on whether or not the AnimationSheet has an anchorPoint defined is also a confusing thing to learn.

But if the old positioning is tossed out completely examples will have to be re-written.

obiot: I would be ok to toss out the old behavior to be honest in the context of a major 3.0.0 release, it was not an optimal and perfect solution anyway, and #309 was aiming at fixing that.

let's wait for @parasyte feedback on this if you are ok :P

Giwayume: If we commit to the new method, the `this.hasTextureAnchorPoint` boolean has no purpose and can be taken out of `AnimationSheet`. It's only there to switch between new and old positioning method at the moment. The `Texture` frame's anchorPoint would still need to be checked if it exists or not, unless it's defaulted in the `Texture` class.

Giwayume:
I have been talking to the owners of TexturePacker. There are some details with how the pivot changes when cropping vs trimming that I didn't include in the code. Will update.

agmcleod:
Question worth asking. Does Shoebox leverage the original code as well? Will this break the shoebox support?

Giwayume:
No, it only adds an anchorPoint if the supporting variables exist. Worst-case scenario for shoebox is they don't get an anchorPoint at the moment and it defaults to current functionality.

I've briefly looked at the shoebox export script and it seems most of the variables are the same if not exactly the same, but I haven't tested it because I didn't want to bother trying to set up legacy Adobe Air on Linux at the time.

Giwayume:
I have tested the shoebox export script. All of the variables are the same as TexturePacker, except the pivot point is missing. So if you export from shoebox the anchorPoint will not be defined in the texture, unless we provide a way to override it.

It's looking more like even if the old positioning is kept, it should be positioning option flag in the entity itself, shouldn't be based on whether or not the texture has an anchor point.

parasyte:
I like that idea! It doesn't seem really useful to have a boolean flag that determines whether to use the anchorPoint. In concept, it's ideal to remove conditional feature support by setting sane defaults when things are missing. I'll run through the patch and let you know if I have any comments. :)

src/renderable/animationsheet.js
```diff
@@ -19,11 +19,12 @@
      * @param {Number} [settings.framewidth] Width of a single frame within the spritesheet
      * @param {Number} [settings.frameheight] Height of a single frame within the spritesheet
```
parasyte, line 20: Don't forget to add the doc comment for the new `settings.anchorPoint` parameter. Just copy-paste this line for the formatting.

Giwayume, line 20: OK. I'm going to update and move the Sprite anchorPoint positioning logic into the Sprite class itself.

parasyte, line 20: That works, too!

Giwayume, line 20: If I draw a Sprite with its anchorPoint at its x and y pos, the debugPanel draws the bounding box incorrectly because it's using this.left (this.pos.x) and this.top (this.pos.y) as the starting point for the bounding rectangle. [debugPanel.js line 200]

Should I override this.left and this.top to take the pixel offset from the anchorPoint into account? I'm not aware of Sprites being used for collision so maybe this isn't much of an issue.

parasyte, line 20: That's a good point. The debug panel is something I've struggled with in the past, as well. I would be cautious about changing the geometry methods (`left` getter et al.) to require information from the parent, like its anchorPoint. We don't have this requirement now, which allows things like inserting a sprite directly into the scene graph without a parent entity.

_on the other hand_ the entity draw method just needs to translate to its own anchorPoint before drawing the child sprite. Then the sprite translates to its anchorPoint as normal. This translation step is clearly missing from the debug panel: https://github.com/melonjs/melonJS/blob/64eeb9996505766a8d3da60339aa5e45d7c18b13/plugins/debug/debugPanel.js#L200

Giwayume, line 20: I'm talking about the Sprite's anchorPoint, not the Entity's (parent's) anchorPoint. The Sprite's left property would be adjusted based off the Sprite's anchorPoint.

spriteanchor

Anyhow, when I modify the left and top to adjust based on the anchor point if it exists, a jasmine test for the droptarget entity fails. It seems the top,left,bottom,right values should be adjusted _only_ for Sprites, not all Renderables.

Making up for this in the debug panel seems hacky to me.

parasyte, line 20: I'm familiar. ;)

Imagine how you might have to modify these lines of code to properly run within `me.Sprite.draw`:

``` javascript
// in this case, the entity's anchor point is in relation to the body
// draw the renderable's anchorPoint at the Entity's anchor point
x = ~~( x + (this.anchorPoint.x * this.body.width) - (this.renderable.anchorPoint.x * this.renderable.width));
y = ~~( y + (this.anchorPoint.y * this.body.height) - (this.renderable.anchorPoint.y * this.renderable.height));
```

Copied directly from your patch, above. :) The issue here is that correctly positioning the sprite relies on the sprite anchorPoint _and_ the entity anchorPoint. Since the entity is already doing this translation, it makes sense that the debugPanel should do it, as well. After all, the debugPanel is just patching `me.Entity.draw` ... Does it make sense?

Giwayume, line 20: Yeah, those lines have changed since the `Sprite` is responsible for drawing itself relative to its own anchor point now. So the problem is not in `me.Entity.draw`, it's in `me.Sprite.draw`.

I'm almost done, you'll see what I'm doing in a couple of minutes.

Giwayume:
OK, so I've defaulted the `Entity` class to draw with the new positioning method. I've added a boolean to the `Entity` class `drawRenderableInBounds` to switch back to the old method.

Because it defaults to the new method, I modified the texture.tps/json/png in the platformer example to set the pivot point at [0.5,1] to match the anchorPoint defined at [0.5,1] in those entities. (modifying these files makes it look like I've touched an extra 250 lines of code =P)

src/renderable/animationsheet.js
Giwayume, line 20: Don't know why this note didn't collapse, but I've updated the doc comment.

parasyte, line 20: This one won't collapse because you haven't changed the line that it points to. It's expected. :smiley: Having a look at the updates, now...

src/renderable/sprite.js
```diff
@@ -262,12 +262,19 @@
...
+            // save context
+            renderer.save();
+            
+            // calculate pixel pos of the anchor point
+            var ax = w * this.anchorPoint.x, ay = h * this.anchorPoint.y;
+            renderer.translate(-ax, -ay);
+            
+            // adjust geometry box position based on anchor point offset
+            this._leftOffset = xpos - ax;
+            this._topOffset = ypos - ay;
```
parasyte: Less is more: https://gist.github.com/parasyte/fa9664cc19824c4ff7b7

I added the same translation into the debugPanel, so the conditional geometry offsets can be removed. The translation is exactly the same as what you do here. It's identical to the method used to draw the hitboxes around the entity; debugPanel needs to do the same translation for those hitboxes as the entity does for its child renderable.

src/renderable/sprite.js
```diff
@@ -262,12 +262,19 @@
...
+            
             if ((this.scaleFlag) || (angle !== 0)) {
-                // save context
-                renderer.save();
 
-                // calculate pixel pos of the anchor point
-                var ax = w * this.anchorPoint.x, ay = h * this.anchorPoint.y;
                 // translate to the defined anchor point
                 renderer.translate(xpos + ax, ypos + ay);
```
parasyte, line 275: Have you tested this with rotate or scale? The new translate line above looks like it will cause this one to translate to the wrong position...

Giwayume, line 275: I have, but now that I think about it I haven't tried a rotation point that isn't the center of the sprite. Checking...

src/renderable/sprite.js
Giwayume: Alright, how do I apply your patch? Or do I just copy it over.

src/renderable/sprite.js
Giwayume, line 275: Yeah, crazy enough it's working. The sprite is rotating around its `anchorPoint`, which is placed at the Entity's `anchorPoint`.

I've found an undesirable behavior, though, that I believe is a pre-existing issue. Normally in game engines scale is applied before rotation so you are rotating a pre-scaled object. However, it looks like it's being done the other way around, and because of this the sprite deforms during rotation when it also has scale applied.

src/renderable/sprite.js
Giwayume: Nevermind, I just copied over the changes and re-committed.

src/renderable/sprite.js
Giwayume, line 275: Placing the `renderer.rotate()` statement before the `renderer.scale()` statement fixes this issue I am referring to. I will commit this small addendum for you to review.

obiot:
wow that looks super good ! i'm really excited by this one :)

two comments I see before we merge this baby :
- we miss a nice entry in the changelog
- do we keep the old `drawRenderableInBounds` options ? i know that we have a history of breaking changes in the framework, but 1st these are changes we discussed since  a long time (so they were wanted), and it's for the 3.0.0 release, a MAJOR one where we can have breaking changes. personally i'm for removing the backward compatible option (plus i hate having a if/else statement in the draw function)

Giwayume:
If you don't think having the old method as an option is useful, I'll remove it.

There's also one more thing I wanted to look at, I'm not entirely sure if the "Crop - Keep Position" option in TexturePacker is working correctly. I know that the other two "Trim" and "Crop - Flush Position" are correct.

Giwayume:
Do you think I should add the old positioning method as a plugin or nah?

parasyte:
I'm happy to remove the optional backward compatibility flag. It should be backward compatible just by passing the body anchorPoint to the sprite constructor, right?

Giwayume:
You made me think for a second, but yes I believe that actually works. Maybe that plugin I made is useless...

Giwayume:
Still waiting on feedback from the TexturePacker team on how "Crop keep position" is supposed to work.

parasyte:
Maybe "backward compatible" isn't the right word, but "to get the old behavior"!

Do you want to wait on the feedback before deciding whether to merge this?

Giwayume:
Yes, the TexturePacker guys have been surprisingly fast to respond.

Giwayume:
Omg they responded to my email like 6 hours ago and I didn't notice... there's definitely a few changes to be made.

Giwayume:
I'm done with the fix.

obiot:
great great work thank you very much @Giwayume , merging now and updating the changelog !

Issue #99, opened 2012-09-05
Separate game logic and drawing updates; allow consistent game speed with variable frame rate


parasyte:
There's a really good writeup on this idea here: http://www.koonsolo.com/news/dewitters-gameloop/

The concept revolves around running the game logic at a constant rate which is *slower* than the frame rate, and allowing the frame rate to run as fast as it possibly can.

The tricky part is properly interpolating object positions in each frame update, between game logic updates. And the article provides some ideas for handling that.

Note: this is very different from the `me.sys.interpolation` feature, which keeps the frame rate and game logic running at the same rate, but attempts to make up for dropped frames by dynamically scaling object velocities.

Giwayume:
What is holding this issue back? Is there a plan to go with the method mentioned in the koonsolo article that involves a constant update fps with drawing interpolation and prediction?

obiot:
yes it was the initial intention, but i don't know why this one kept falling behind in terms of priority and we never ended up doing it :(

Giwayume:
Currently it looks like both of the first to methods mentioned in the article are implemented, with a switch between the two by setting my.sys.interpolation.

So to implement a "prediction" solution, I'm guessing the API would look like:
```
me.sys.fps = 25; // Logic fps 
me.sys.interpolation = true; // I'm guessing this is optional.
me.sys.framePrediction = true; // New variable, whatever it's called. Turns on prediction.
```
For the prediction code, it looks like only `me.Container` and `me.Sprite` will have to change to utilize it? Are there other renderables that affect positioning frame-by-frame? It would be another IF statement in each drawing call.

I something similar to MAX_FRAMESKIP mentioned in the article implemented today?

obiot:
for me the only place this should be implemented is here :
https://github.com/melonjs/melonJS/blob/master/src/game.js#L248-L269
which is the main loop "update" function 

or here :
https://github.com/melonjs/melonJS/blob/master/src/state/state.js#L190-L199
as this is the callback used by the RaF function

but other objects should stay as agnostic as possible from this (in my opinion) and rely only on the given `dt` information passed to the update function

Giwayume:
I don't think it is possible to implement at that high of a level.

The idea is you're limiting the actual update logic to 25 fps, so each `Renderable` needs to "guess" its position based on the last "update" frame. So this is an extra calculation per `Renderable`.

obiot:
honestly, my personal intention was rather to fix the timestep as indicated in the following article:
[...]

Issue #16, opened 2011-10-02
Falling Bug


kosinix:
Hi,

First, this is a great javascript engine. I ported an old platform game I made in flash. However, there is one critical bug when falling from high places. For example, when jumping from high platform the hero goes through a solid tile. My theory is this has something to do with the acceleration of the fall greater than the height of the solid tile, as I encountered this kind of bug with my flash engine from long time ago. Just hoping you could check this bug, that would be really cool.

thanks! Really nice game engine.

obiot:
ok, I'm closing this one for the following reasons :
- the problem was only visible in earlier version of melonJS, and the cap to `maxVel` is "hiding" the issue
- the ticket is redundant with other tickets were we discussed about having a better physic implementation : #235 #285 #588 , in the sense that the later 3 will not have such issue.

PR #718, opened 2015-08-01
Suggested solution for Issue #99

Giwayume:
Here's how it works:

Game loop automatically calls the update function once or multiple times based on `me.sys.updatesPerSecond` as the update rate.

For the user to implement frame prediction in their game, `me.sys.updatesPerSecond` must be lower than `me.sys.fps`. When this happens, draw events will be called as often as `me.sys.fps` and update events will be called at `me.sys.updatesPerSecond`.

Frame prediction demo added. From what I can tell, having a high `me.sys.updatesPerSecond` works a little smoother than implementing frame prediction.

Setting `me.sys.interpolation` to true disables all of the above, and works how it does today.


examples/frame_prediction/js/entities/entities.js
```diff
@@ -0,0 +1,76 @@
...
+/************************************************************************************/
+game.FramePredictionEntity = me.Entity.extend({
+   
+    init: function(x, y, settings) {
+        this._super(me.Entity, "init", [x, y , settings]);
+    },
+    
+    draw: function(renderer) {
+        // Predict the next position to draw at based on velocity
+        var updateInterval = (me.timer.getTime() - me.timer.lastUpdate) / ( (1000/me.sys.updatesPerSecond) );
```
parasyte, line 14: Right now the draw method is passed two arguments:
1. Renderer reference
2. Visible rectangle

I'm not sure how useful the rectangle is. For prediction purposes it would be ideal to just pass a delta time as the second argument, as done with the update method. @obiot @agmcleod Any opinions on this?

src/state/state.js
```diff
@@ -188,14 +188,14 @@
...
-            // render all game objects
-            me.game.draw();
             // schedule the next frame update
             if (_animFrameId !== -1) {
                 _animFrameId = window.requestAnimationFrame(_renderFrame);
             }
+            // update all game objects
+            me.game.update(time);
+            // render all game objects
+            me.game.draw();
```
parasyte, line 194: What, (if anything) is affected by moving this code?

Giwayume, line 194: It's supposed to create a smoother animation when you draw after making the frame request, because you're making the request sooner. I've seen quite a few articles mention this. (this one looks like it's talking about the setTimeout polyfill, though)

http://www.paulirish.com/2011/requestanimationframe-for-smart-animating/

examples/frame_prediction/js/entities/entities.js
Giwayume, line 14: Note the interval is calculated based on the time from the last update call, _not_ the time from the last requestAnimationFrame. A delta time based on your fps as opposed to the updatesPerSecond is not useful for frame prediction between updates.

src/state/state.js
Giwayume, line 194: I'm not sure of the validity of this suggestion, though. I've seen no tests to prove that it actually creates a smoother animation. It's probably all based on perception and assumptions.

parasyte, line 194: This code change cannot change the result, because JavaScript is single-threaded and _uninterruptible_. The only way to get smoother animations is to make sure the `draw()` _and_ `update()` methods complete within the ~16.667ms timeframe (for 60FPS). The order of operations cannot speed it up. ;)

Giwayume, line 194: I get that, but requestAnimationFrame is an asynchronous callback. Once all of the single-threaded code for one execution of the loop is done, there's a period of time where the code is either waiting, doing nothing, for the callback to come _or_ the callback is already resolved and executes immediately after the synchronous code for update() and draw() have completed. So there's a variable here of how long it takes for the asynchronous callback to resolve. That's the logic, in theory.

And yes, it would not apply to 60fps, but any fps lower than that is fair game.

Giwayume, line 194: Here's a longer discussion article for this issue involving another Javascript game engine:
http://impactjs.com/forums/impact-engine/misuse-of-requestanimationframe/page/1

(pretty much a bunch of back and forth and "we're not sure")

parasyte, line 194: It's asynchronous by the nature of video-sync. :) The callback is queued for the next available frame. In this context, "available" just means that no other code is currently running.

Imagine you had a lot of complex AI and particles, and your frame processing was taking 18ms. With an RAF callback triggered at T+0, the frame processing goes beyond the 16.667ms timeframe you have for T+1. So the UA ends up dropping a frame, and the next RAF callback fires at T+2, regardless of what point during that 18ms processing time the RAF was scheduled; it could have happened before or after T+1, and the result will still be the same. It drops a frame at T+1 because JavaScript is currently running at that time.

This is what makes it better than a timer (or interval) for animations; the operation of RAF is deterministic.

Giwayume: Regardless of the magic the browser creates with requestAnimationFrame, this is an actual issue when using setTimeout.

Giwayume, line 194: Regardless of the magic the browser creates with requestAnimationFrame, this is an actual issue when using setTimeout.

Here's the case:

Call setTimeout (16ms) -> Code executes for (15ms) -> setTimeout resolves 1ms later -> draw!

Code executes for (15ms) -> Call setTimeout (16ms) -> setTimeout resolves 16ms later -> missed a frame!

parasyte, line 194: You're talking about the polyfill we use. In that case, it's better to run the polyfill version of RAF at the end, where it can get a more accurate delta time that properly handles frame dropping. Here's another example case to consider:
- Call `setTimeout(16ms)`
- Code runs for 24ms
- Timeout callback fires >8ms late

When this happens once, it doesn't matter too much. When it happens hundreds of times in a row, the late callbacks add up, and the game actually runs slower. To fix this, polyfill RAF at the end:
- Record current time in polyfill RAF callback
- Code runs in 24ms
- Call `setTimeout(delta)` where `delta` is computed modulus (1000 / 60);

``` javascript
var step = 1000 / 60;
var delta = step - ((currentTime - lastTime) % step);
```

This replicates the magic of RAF.

Giwayume, line 194: Alright lol I see you've got this whole thing thought out.

parasyte, line 194: Note that the polyfill doesn't handle frame dropping as described: https://github.com/melonjs/melonJS/blob/ba679b15a0806d91c1a9ea370f1ae312f5c42643/src/state/state.js#L86

parasyte, line 194: It's copied from Paul Irish's blog, which you linked earlier in the discussion.

examples/frame_prediction/js/entities/entities.js
obiot, line 14: the second argument just basically contains the visible region to redraw. it used to be more extensively used by the former dirty rectangle algorithm (if you remember), but now we just basically pass the viewport (me.TMXLayer is most probably the one using it the most).

else for prediction purpose, why would the renderer not store the last time information when the last frame was rendered ? 

and as i was suggesting it somewhere else, what about we move the RaF code/callback and main loop into our renderer object ? it then can itself do all this kind of operations

Giwayume, line 14: For each draw call, the Entity is starting out in the state of time of when it was last updated.

So for the prediction demo, we have 4 draws for every 1 update. Each time you draw, the entity's stored information (x, y, rotation, etc) is still stuck in time 0-3 frames ago. So you need to know the timestamp from when this information was last updated to make a prediction, not the timestamp from the last frame.

This is why I added `me.timer.lastUpdate = time;` in the update loop.

parasyte, line 14: The rendered should be agnostic of timing information, because their purpose is drawing one frame at a time. Not animating.

And I made reference to the multiple viewport ticket previously as an example. The renderer API we have now is pretty much what's necessary to build multiple viewports. It would have to get really hacks if each renderer was doing RAF. Not cool.

The TMX layer classes are interesting, since they can optimize drawing based on the visible rectangle. It sounds like that needs to change for multiple TMX support, too ;p

Anyway, passing the delta time for draws would be ideal for predictive positioning.

Giwayume:
Shoot there's one thing I'm forgetting,
`me.game.viewport.follow(this, me.game.viewport.AXIS.BOTH);`

How would the user modify this function to implement predictive positioning?

examples/frame_prediction/js/entities/entities.js
Giwayume, line 14: You guys are saying the rectangle argument is still used, so pass the update delta time as a third argument?

parasyte, line 14: I'm trying to work out in my head the best API. Certainly open to suggestions, but I feel the rectangle is going to be less useful than a time delta for animations in the draw method. That would prioritize the time delta as the second parameter, and an optional rectangle as the third.

Another point for passing the time delta to the draw methods is that all animation code (animation sheet, tile animations, ...) can be moved into the draw method, where it really belongs. Without this change, running updates at a frequency different from drawing would cause animation frames to drop (low frequency updates), or waste CPU (high frequency updates).

parasyte:
For viewport follow, it makes a good case for implementing predictive positioning within the engine itself, right? melonJS internals can use the same velocity interpolation for entity positioning and viewport positioning. It also seems "nicer" this way than deferring the feature to user code.

Giwayume:
It would seem so, but then again the user can still call `me.game.viewport.move();` I suppose and calculate that out manually.

If it's done in the engine, more than just the game loop needs to be touched. Also there's things like rotational velocity which we don't have a variable for today... so if the user wanted a smooth rotation animation they'd still have to do some manual coding.

parasyte:
We also have full-featured tween support, which is better suited to interpolation. It works with rotation, scaling, and translation.

examples/frame_prediction/js/entities/entities.js
obiot, line 14: the good thing is that the second paramater is actually not documented, and as I was saying earlier only used by the `me.TMXLayer` object, so we are really free to do whatever we prefer with it (adding parameters will probably break a few games though, for all the people that are redefining the draw method and calling the parent one)

Renderable API (note the typo on the argument name) :
http://melonjs.github.io/docs/me.Renderable.html#draw

Giwayume, line 14: The docs/examples should probably recommend this method of calling supers so that code breaking doesn't happen:

```
this._super(me.Entity, 'draw', arguments);
```

Giwayume:
Since tweens use the update function, they won't be useful for animation for frame prediction. Should the tween update function change to draw?

Giwayume:
Also, tweens actually modify the properties they're attached to. For frame prediction, the draw logic should not interfere with the update logic.

Giwayume:
Implementing in-engine frame prediction looks like it's going to touch a *lot* of stuff.

Can we agree on the implementation of the stable update solution this code provides (calling update multiple times per draw), then move frame prediction logic from the user's hands to in-engine as a separate merge?

I feel like the scope the frame prediction fix is getting too large, and moving frame prediction in-engine should be handled as a separate entity. Unless you guys are OK with huge pulls.

obiot:
To follow the agile philosophy (that i dont't agree with btw), it's always better to break down a huge task into smaller size task (so that it is easier to plan/organize them over several iteration/versions). So yes i would personnally agree to it :)

parasyte:
Yes, predictive positioning is different from update accuracy. (For reference see: https://github.com/melonjs/melonJS/issues/99#issuecomment-126942149)

Giwayume:
I think technically #99 is solved with this code, a different issue should be opened for in-engine predictive positioning. There's a lot to talk about there.

Are there any issues with this implementation of multiple update calls?

src/game.js
```diff
@@ -209,6 +218,14 @@
             // reset the frame counter
             frameCounter = 0;
             frameRate = ~~(0.5 + 60 / me.sys.fps);
+            
+            // set step size based on the updatesPerSecond
+            stepSize = (1000 / Math.min(me.sys.updatesPerSecond, me.sys.fps));
```
parasyte: This prevents `updatesPerSecond` from being larger than `fps`. Because the `stepSize` is capped to `fps`, it makes the new while loop in `me.game.update` useless; `updateDelta` can never be higher than ~16.6667, which is roughly what `me.timer.getDelta()` returns during RAF.

In other words, this use of `Math.min` limits the while loop to acting as an if-statement; it will only execute the body zero or one time.

Giwayume: From what I'm aware, `me.sys.fps` is the _target_ fps, not actual fps. So updatesPerSecond will never run faster than the target display fps, which is the intention. (is there a need to run your update calculations faster than your targeted (max) display fps?)

The intent is to always have your update calculations run at the target updatesPerSecond (60fps) even when the draw fps drops down to say, 20.

From my tests running a game at 20/60 fps, this code is working. The character moves at the same speed when the game runs at 20/60 fps as it does when the game runs at 60/60 fps.

parasyte: Let's try it.
- `me.sys.updatesPerSecond = 20;`
- `me.sys.fps = 60;`
- `stepSize = (1000 / 20);` stepSize will be 50.
- in `update`:
  - `accumulator += me.timer.getDelta();` accumulator =~ 16.6667
  - `updateDelta = stepSize` ... 50
  - `while (accumulator >= stepSize)` ... nope, not yet! Updates will happen later (body executes zero times)
  - later ... (exactly 3 frames)
  - `accumulator += me.timer.getDelta();` accumulator =~ 50.0001
  - `while (accumulator >= stepSize)` true!
    - `accumulator -= stepSize;` accumulator is now ~0.0001. The while loop runs only once.

Now let's rewind and run with `updatesPerSecond > fps`, just for giggles...
- `me.sys.updatesPerSecond = 200;`
- `me.sys.fps = 60;`
- `stepSize = (1000 / 60);` stepSize will be ~16.6667 (this is where capping to `fps` happens)
- in `update`:
  - `accumulator += me.timer.getDelta();` accumulator =~ 16.6667
  - `updateDelta = stepSize` ... ~16.6667
  - `while (accumulator >= stepSize)` true!
    - `accumulator -= stepSize;` accumulator is now ~0.0001. The while loop runs only once.

Hmmm...

Giwayume: Yeah, my point with the last comment was:

```
me.sys.updatesPerSecond = 200;
me.sys.fps = 60;
```

Do you _really_ need to update faster than you can possibly draw? Why? I figured not, that's why I put that in, so it becomes:

```
me.sys.updatesPerSecond = 60;
me.sys.fps = 60;
```

Giwayume: Even something more realistic like
```
me.sys.updatesPerSecond = 60;
me.sys.fps = 30;
```
Let's say you're doing this for more accurate collision checks at 60fps. If so, why limit your drawing to 30fps? Why not let the browser draw at 60fps if it can? I'm just not seeing the use case here.

Giwayume: Remember my comment https://github.com/melonjs/melonJS/issues/99#issuecomment-126951678?

That's how I coded it.

`me.sys.fps`  // Max fps (draw **and update**)

parasyte: This comment: https://github.com/melonjs/melonJS/issues/99#issuecomment-126942149 For collision accuracy (e.g resolving issue #16) the collision detection is done at a higher frequency than drawing, giving more intermediate states, preventing high-velocity entities from skipping collisions that "obviously" should occur.

That's why you would want to run `update` multiple times per frame.

However, you still haven't addressed my comment that the while-statement never loops? It only executes zero or one time, which is the behavior of an if-statement.

Giwayume: Assume one update takes 2ms and one draw takes 30ms.

`me.timer.getDelta()` will return at least 32.

You have made the assumption above that the game is running at a perfect 60fps in the line:
`accumulator += me.timer.getDelta();` accumulator =~ 16.6667

In which case if it is running at 60fps it is correct that the while loop executes only once. It's meant to execute multiple times when it takes longer to draw.

parasyte: Got it, that's the case I didn't consider.

What happens if `update` always takes longer than 16.6667ms? Do we get an infinite regress of slower updates? There's no way it could possibly catch up...

Giwayume: No, because stepSize is a constant. The game simply runs slower.

Giwayume: I'll remove `Math.min(me.sys.updatesPerSecond, me.sys.fps)`, but we're still limited to 60 updatesPerSecond max because of RAF. 

EDIT: Actually, not necessarily. My bad.

src/game.js
```diff
@@ -253,18 +271,30 @@
...
-                isDirty = api.world.update(me.timer.getDelta()) || isDirty;
-
-                // update the camera/viewport
-                isDirty = api.viewport.update(me.timer.getDelta()) || isDirty;
+                
+                accumulator += me.timer.getDelta();
+                
+                updateDelta = (me.sys.interpolation) ? me.timer.getDelta() : stepSize;
+    
+                while (accumulator >= stepSize || me.sys.interpolation) {
```
parasyte: Picking up where we left off with the other thread, this loop will cause problems for updates that take too long. In that case you want a bigger `stepSize` and fewer updates.

Giwayume: I think for the most part that's a game developer problem. When they choose a number for `updatesPerSecond` it should be realistic in terms of how heavy their update logic is.

I guess for this problem, put a max on the accumulator size. Maybe 4x the stepSize.

parasyte: Compounding accumulators are more than the developer's problem. With this while loop, it will tank the engine.

Let's take a hypothetical update that runs for 25ms. Of course, this causes one frame (draw) to be dropped. The dropped frame causes the following update to be run twice, to "catch up" the time lost from the dropped frame. The loop runs for `25ms * 2`, dropping 2 additional frames. The next update will try to catch up again, running for `25ms * 3`, ad infinitum.

Long updates will exhibit polynomial growth in time. The cap that you just added means that the polynomial growth is capped, but at that point the game slows down, canceling any net effect from attempting to catch up.

`me.sys.interpolation` approaches this issue from the other angle; allow frames to drop, and scale velocities et al to compensate.

Giwayume: What do you suggest? Measuring update and draw times and comparing them?

Giwayume: I was able to reproduce this polynomial growth behavior, but only after I boosted updatesPerSecond all the way up to 100000. =P

I'm not thinking it's going to be a big issue.

parasyte: :laughing: You just need a lot of entities with a lot of AI stuff. It's actually not too difficult to get long update times. Especially on slow devices like my awful iPhone 4S ...

obiot:
on a side (but related) note, the tween code most probably needs a small update to also take in consideration the given `dt` passed by the update function :
https://github.com/melonjs/melonJS/blob/master/src/vendors/tween.js#L312
as for now, it just used the `me.timer.getTime();` function, instead of rather using a accumulator.

Giwayume:
Is tweening ever used for physics?

obiot:
not in melonJS, but games like Flappy bird is using it to make the bird fly in smooth way :P

Giwayume:
Changed tween.js to use the update dt in relation to me.timer.lastUpdate. Also fixed the anchor points in the whack-a-mole example.

src/game.js
Giwayume: I've tried to limit the accumulator based on how much time a single update takes vs the step size, but that just results in a lot of on and off lag.

obiot:
Oh man, i just did the same a few hours ago for the whack-a-mole example, and it then now results in conflict on this branch :(
https://github.com/melonjs/melonJS/commit/3c61475a71259563592529013e57567e0412689a
Can u maybe rollback these changes on your side?

Giwayume:
How do I do that? I'm not all that familiar with git.

obiot:
In theory you could rollback on your branch, then redo a pull request? @parasyte ?

Giwayume:
I tried checking out the files from the previous commit, the committing them. It looks like that worked.

obiot:
:+1:

Giwayume:
Can anyone think of a better way to handle slower computers than `accumulatorMax`? If not, anything else?

Giwayume:
I have looked around at different articles about the multiple update solution, but none of them have addressed the problem when your updates take longer than the delta step size.

obiot:
I thought that using `accumulatorMax` was the conclusion of your discussion yesterday between you and @parasyte, is this not the case ?

Giwayume:
"The cap that you just added means that the polynomial growth is capped, but at that point the game slows down, canceling any net effect from attempting to catch up."

A true statement. I'm not sure if that was an indication to look for a better solution or what.

obiot:
i'm afraid it's better to wait for Jason to be back on this one !

parasyte:
It feels like the best way around that edge case is something like `updateDelta = Math.max(me.timer.getDelta(), stepSize);` ... But really this just popped into my head! I haven't had any time to think it through.

Giwayume:
Hmm... doing that speeds up the animation on slower machines for some reason? 

Also, It's not taking into account update vs draw time. I've tested it on a slow laptop that takes 2ms to update and 70ms to draw. Obviously 2ms is well below the stepSize needed to update multiple times but 72ms is not...

parasyte:
That's a good point. Separate time deltas for update and draw? For now, `getDelta` is returning the delta for the entire frame.

Giwayume:
Following your logic above, I used an average update delta instead of the timer delta. Only thing that bothers me is I'm not sure how costly `window.performance.now()` is to run. From testing, it seems to work OK.

Giwayume:
Yeah, one more thing. Animations are slowing down at a step size higher than the fps, and speeding up when updates can't match the fps.

Giwayume:
I guess this is expected behavior, though.

parasyte:
Is that because you used an average, instead of the actual delta per frame? As you know, it only needs to use a larger time step if the update on the previous frame took longer than `updateDelta`ms.

Giwayume:
Looks like the animation slowdown happens fire a different reason than the speedup. The speedup was because of updateDelta increasing due to the `Math.max(stepSize, updateAverageDelta)`.
The slowdown was because I put updatesPerSecond so high that the delta timestamp calculations were too small.

I put in an average of the last 3 frames because otherwise an fps drop drop spike would pause the multiple updates if it landed on the very last update call before the draw call, which would limit the next set of update calls and create a lag spike.

Giwayume:
I think I'm done messing with this.
I've tested before and after performance on 3 computers in the platformer demo, 60 updatesPerSecond:

Slow laptop:
* Before: 43.47 updatesPerSecond, 43.42 drawsPerSecond
* After: 59.27 updatesPerSecond, 44.06 drawsPerSecond

Mid-range laptop:
* Before: 55.16 updatesPerSecond, 55.10 drawsPerSecond
* After: 59.83 updatesPerSecond, 57.69 drawsPerSecond

Top-of-the-line gaming desktop:
* Before: 59.80 updatesPerSecond, 59.73 drawsPerSecond
* After: 59.95 updatesPerSecond, 59.69 drawsPerSecond

I was surprised to see that draws per second increased in the lower-end machines, but that's probably statistically irrelevant and would have went away if I averaged multiple performance tests.

parasyte:
Seems good to me! Merging now. We can further tune it as necessary.

Issue #637, opened 2015-01-18
[WebGL] Reconcile the global matrix


parasyte:
This ticket will pick up where #591 leaves off. The goal here is to remove dependence on the global matrix, as much as possible. It will allow us to put almost every matrix transformation entirely within the GPU, where it can take advantage of highly parallelized vector ALUs that we don't have access to in JavaScript.

To start, recall that 3D rendering typically uses three matrices; called "model matrix", "view matrix", and "projection matrix". Working backwards, melonJS needs the projection matrix to transform our pixel coordinates to WebGL cube space coordinates using orthographic projection; and a view matrix is used for nested containers and other object that need relative drawing operations. The model matrix is unnecessary because the majority of entities drawn in melonJS are quads (exactly two triangles). A model matrix is more useful with large meshes, which have a greater number of vertices.

The scaling and rotation values can be streamed to the GPU as part of the vertex attribute array. These can be packed into 2 floats (a 3x3 matrix requires 6 floats for full precision) assuming 16 bits for each of the scaling coordinates. Additionally, the rotation angle and texture index can be packed into the same float. The anchor point will also be required, which is a 3rd float. These attributes can be applied within the vertex shader, since the order of operations is static; translation can be added to the destination coordinates before calling `drawImage`.

Entities that need a model matrix can implement one with a custom compositor and shader. Spine is an example that will benefit from a model matrix.

To reconcile the global matrix, we need to remove most of the matrix transformation usage. `me.Sprite` and `me.Particle` make heavy use of the global matrix right now. Here is the last proposal for a replacement: https://github.com/melonjs/melonJS/issues/591#issuecomment-68618550 The important part is that the first argument provides information about things like scaling and rotation for each `drawImage` call. The function signature will have to change to accept a texture region instead of coordinates, as an additional optimization. See: https://github.com/melonjs/melonJS/commit/e785724ac5439f27e43a3a9c42a5eaa082bff1f2#commitcomment-9329151

With that part done, each view matrix transformation will be treated as change in WebGL state; it will be sent to the GPU as a uniform variable, so changes to the matrix require flushing the compositor. This will be pretty efficient, assuming view matrix transformations only occur on `me.Container` boundaries.

Additional TODO items:
- [ ] Implement fonts (hard), lines (easy), stroke (hard)
- [ ] Create a packet inspector for debugging
- [x] Enable mipmapping on power-of-two textures
- [ ] Pack multiple attributes into each float
[...]

Issue #335, opened 2013-12-18
Move scaling, rotation, alpha, etc to me.Renderable


parasyte:
For discussion: I think the me.Renderable base class should be responsible for all drawing, including scaling, rotation, and opacity/alpha blending.

I'm indifferent about keeping backward compatibility by continuing to support these properties on ObjectEntity.

obiot:
considering that we did not really agree here on how to properly do it, I'm moving it to 4.0.0 as well (won't likely happen on 3.0.0)

obiot:
hi guys, 

I was looking back at this one and i really wanted to propose the following to address this ticket :

* (Function) `setTransform(scaleX, skewX,  skewY, scaleY, posX, posY)` : which provides a convenience function that mimic the standard [function](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/setTransform)
* (Function) `transform(scaleX, skewX,  skewY, scaleY, posX, posY)`, same as above, but add to the already defined transformations instead of resetting them.
* (Function) `scale(x,[y])` (from `me.Renderable) 
* (Function) `rotate(angle)` : that replaces the properties `angle` we had in `me.Renderable
* (Function) `hasTransform()` : returns true if any transform was applied to the object
* (Function) `getTransform()` :  used by renderers to apply transform defined by a renderable

All related properties would then be stored in a mat2d object, allowing then to easily pass back the same object to the [Canvas Renderer](https://github.com/melonjs/melonJS/blob/master/src/video/canvas/canvas_renderer.js#L534) or [WebGL Renderer](https://github.com/melonjs/melonJS/blob/master/src/video/webgl/webgl_renderer.js#L675) `transform` function.

with that, any class inheriting from me.Renderable, just have to focus on drawing whatever they need without having to call something else or a parent function or whatever. As the main loop in the [container object](https://github.com/melonjs/melonJS/blob/master/src/renderable/container.js#L789) would then be in charge of applying any defined transforms before calling the object draw function, so something like that  :
````javascript
           for (var i = this.children.length, obj; i--, (obj = this.children[i]);) {
                hasTranform = obj.hasTransform() || obj.floating;
                if ((obj.inViewport || isFloating) && obj.isRenderable) {
                    if (hasTranform) {
                        // translate to screen coordinates
                        renderer.save();
                        renderer.resetTransform();
                    }

                    // draw the object
                    if (hasTranform) {
                        renderer.transform(obj.getTransform());
                    }
[...]

PR #727, opened 2015-08-20
Fix for TexturePacker/ShoeBox rotate property

Giwayume:
This appears to resolve the issue from my side. Tested rotation and scaling with a TexturePacker animation.

### DIFF
--- a/src/renderable/sprite.js
+++ b/src/renderable/sprite.js
@@ -279,14 +279,17 @@
                 if (this.scaleFlag) {
                     renderer.scale(this._scale.x, this._scale.y);
                 }
-                renderer.translate(-(xpos+ax), -(ypos+ay));
                 // remove image's TexturePacker/ShoeBox rotation
                 if (this._sourceAngle !== 0) {
-                    renderer.translate(this.height/2, this.width/2);
+                    renderer.translate((this.height/2)-(xpos+ax), (this.width/2)-(ypos+ay));
                     renderer.rotate(this._sourceAngle);
-                    renderer.translate(-this.height/2, -this.width/2);
+                    xpos -= this.height/2;
+                    ypos -= this.width/2;
                     w = this.height;
                     h = this.width;
+                } else {
+                    xpos = -ax;
+                    ypos = -ay;
                 }
             }
 
--- a/src/renderable/sprite.js
+++ b/src/renderable/sprite.js
@@ -266,7 +266,8 @@
             
             // calculate pixel pos of the anchor point
             var ax = w * this.anchorPoint.x, ay = h * this.anchorPoint.y;
-            renderer.translate(-ax, -ay);
+            xpos -= ax;
+            ypos -= ay;
             
             if ((this.scaleFlag) || (this.angle !== 0) || (this._sourceAngle !== 0)) {
                 // translate to the defined anchor point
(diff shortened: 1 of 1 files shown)


src/renderable/sprite.js
```diff
@@ -269,30 +268,25 @@
...
                     renderer.scale(this._scale.x, this._scale.y);
                 }
-
+                renderer.translate(-(xpos+ax), -(ypos+ay));
+                // remove image's TexturePacker/ShoeBox rotation
                 if (this._sourceAngle !== 0) {
-                    // swap w and h for rotated source images
+                    renderer.translate(this.height/2, this.width/2);
+                    renderer.rotate(this._sourceAngle);
+                    renderer.translate(-this.height/2, -this.width/2);
```
parasyte: Wow, I counted five translations in all. How many of these can we drop?

Giwayume: We're doing two separate rotations. In order to rotate around a certain point, you need to translate the canvas to where the rotation point is at the origin. That's 2 minimum. The position resetting translations may be removable. Got any ideas?

parasyte: Two is certainly better than five. Even three is better than five, if you want to avoid a full save/restore between the rotations. Translate is commutative. :smiley:

src/renderable/sprite.js
```diff
@@ -269,28 +268,26 @@
...
                     renderer.scale(this._scale.x, this._scale.y);
                 }
-
+                // remove image's TexturePacker/ShoeBox rotation
                 if (this._sourceAngle !== 0) {
-                    // swap w and h for rotated source images
+                    renderer.translate((this.height/2)-(xpos+ax), (this.width/2)-(ypos+ay));
+                    renderer.rotate(this._sourceAngle);
+                    xpos -= this.height/2;
+                    ypos -= this.width/2;
```
parasyte: I don't think the rotation is supposed to be performed around the center. Assume an image with a 2:1 or 1:2 ratio; rotating around the center will change the position of the image, relative to its anchor point.

Giwayume: I tried rotating it around the corners first, but that wasn't working at all. This method seemed to be accurate at a glance, some of my rotated frames are greater than 2:1 ratio. I'll try something more drastic.

Giwayume:
Not yet fixed. Current code only works when rotated frame's original width is greater than height.

obiot:
i really don't know how, but somehow it would be great to have a testunit for this, as the day we replace all this code some matrix operations, it's gonna be fun to track down any regression :P

Giwayume:
Only way I can think of testing this is drawing a sprite to a canvas and seeing if the pixels match what's expected.

parasyte:
@obiot I don't think matrices will solve this. A matrix is just a method of combining multiple transformations into a single state. You still have to rotate the matrix twice (minimum) to handle both rotations here.

obiot:
@parasyte i more see a matrix useful to "store" the current transform status of an object, instead of having all these properties, and it can also be applied in one easy shot before rendering the object, but this is another discussion we already had in #335 :)

parasyte:
On the other hand, that "one easy shot" isn't always easy with WebGL: https://github.com/melonjs/melonJS/issues/637#issue-54693623 Anyway, this is pretty off-topic here.

Giwayume:
I can't break this code in its current state. Looks like it's done.

parasyte:
LGTM! Thanks again for the contribution!

Issue #730, opened 2015-08-23
me.ParticleEmitter z pos always displays on top


Giwayume:
Look into this and also possibly change the emitter to extend me.Container instead of owning a container as a property.

Giwayume:
It looks like the reason the emitter is not the container is because they would share the pos variable otherwise. The container needs to be at position 0,0 while the emitter's position controls the particle spawn point.

I think a quick and dirty solution would be to have the emitter automatically place the container instead of leaving it up to the user.

PR #731, opened 2015-08-24
ParticleEmitter Z sorting and container

Giwayume:
Turns out my issue with #730, I was trying to set pos.z and it was expecting plain-old z. This fixes that, also the ParticleEmitter handles adding and removing its particle container instead of relying on the user to do so.

### DIFF
--- a/examples/particles/js/editor/controls.js
+++ b/examples/particles/js/editor/controls.js
@@ -32,7 +32,6 @@
 
         clear : function() {
             for ( var emitters = this.emitters, i = emitters.length, obj; i--, (obj = emitters[i]);) {
-                me.game.world.removeChild(obj.container);
                 me.game.world.removeChild(obj);
             }
             this.emitters.length = 0;
@@ -46,7 +45,6 @@
             emitter.name = "emitter" + me.utils.createGUID();
             emitter.z = 10;
             me.game.world.addChild(emitter);
-            me.game.world.addChild(emitter.container);
             emitter.streamParticles();
             this.addEmitter(emitter);
             this.selectEmitter(emitter);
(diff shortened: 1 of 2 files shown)


obiot:
Turns out it was actuallt a nasty regression, thanks for the patch!

parasyte:
That patch looks fine to me, for what it is. The whole particle system could use some love.

Giwayume:
I like the way the ParticleEmitter API works. Maybe the way the classes are organized internally in the engine isn't the best, but it's simple to set-up from a game developer perspective.

obiot:
Let's merge it then :)

examples/particles/js/editor/controls.js
```diff
@@ -472,7 +469,6 @@
                 code.push("emitter.name = '" + emitter.name + "';");
                 code.push("emitter.z = " + emitter.z + ";");
```
Giwayume, line 470: I forgot to change this part of the example to pos.z...

obiot, line 470: what should be it be updated to ? here above you also refer to `emiter.z` (line 46/47)

examples/particles/js/editor/controls.js
```diff
@@ -46,7 +45,6 @@
             emitter.name = "emitter" + me.utils.createGUID();
             emitter.z = 10;
```
obiot, line 46: emitter.pos.z ?

Giwayume, line 46: Yes that is correct.

obiot, line 46: @Giwayume so no need to change the example part below, correct ?

Giwayume, line 46: I don't know what you mean by "below", but all references to emitter.z just won't work. They should all be emitter.pos.z... I didn't look through the example too hard when I made the change. Was mostly testing it out in my own game.

obiot, line 46: yes so that's what i was saying/asking, based on thsi comment :
https://github.com/melonjs/melonJS/pull/731/files#r37757771

Issue #736, opened 2015-09-05
Update debugPanel plugin (and debugParticlePanel?) to use me.Object.extend


parasyte:
Reported on the forum: https://groups.google.com/d/msg/melonjs/fAtFFe2wvCo/f4szE5o1BwAJ

obiot:
I'm confused about this one, as it's already the case :
https://github.com/melonjs/melonJS/blob/master/plugins/debug/debugPanel.js#L24
https://github.com/melonjs/melonJS/blob/master/plugins/debug/debugPanel.js#L46
https://github.com/melonjs/melonJS/blob/master/plugins/debug/particleDebugPanel.js#L19

and the current version of debugPanel is working just fine with the platformer, something else i'm missing?

parasyte:
Hmmmm, nope! LGTM. User error.

PR #738, opened 2015-09-05
Updated platformer example so that it exposes the parallax bug.

Giwayume:


Giwayume:
As far as the bug goes, the anchor point is arriving to the code as a string "{"x":0.5,"y":0.5}", not a json object.

parasyte:
Use `json:{"x":0.5,"y":0.5}` in the Tiled property value. The `json:` prefix is important.

Giwayume:
One question. If the background image is the exact width/height of the room and the anchorPoint ends up being <Infinity,Infinity>, is that going to break anything?

parasyte:
It shouldn't become Infinity. It could become NaN due to a divide by 0, but that's why the patch in #736 has `|| 0`.

Giwayume:
If I do 1/0 in the inspector of Chrome or Firefox I get Infinity.

Giwayume:
http://stackoverflow.com/questions/18838301/in-javascript-why-does-zero-divided-by-zero-return-nan-but-any-other-divided-b

parasyte:
Oh, right >_> Thanks, JavaScript.

```javascript
Math.min(x, 0) || 0;
```

parasyte:
Fixed by 83e4ae010f90a0e21cefcbca45e6a216f1cc88a0

Giwayume:
Do you want the parallax clouds in the platformer example or should I just toss this?

parasyte:
I added a `platformer2` example based on this one that has parallax scrolling backgrounds.

PR #743, opened 2015-09-14
adds a proper init method to reset the vars, and ensure everything is

agmcleod:
good with the life cycle methods

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -111,9 +111,6 @@
         "afterEach"             : false,
         "spyOn"                 : false,
         "xdescribe"             : false,
-        "xit"                   : false,
-
-        // objects
-        "Map"                   : false
+        "xit"                   : false
     }
 }
--- a/.jshintrc
+++ b/.jshintrc
@@ -111,6 +111,9 @@
         "afterEach"             : false,
         "spyOn"                 : false,
         "xdescribe"             : false,
-        "xit"                   : false
+        "xit"                   : false,
+
+        // objects
+        "Map"                   : false
     }
 }
--- a/src/vendors/tween.js
+++ b/src/vendors/tween.js
@@ -62,7 +62,7 @@
             }
         };
 
-        function init (object) {
+        this.setProperties = function (object) {
             _object = object;
             _valuesStart = {};
             _valuesEnd = {};
@@ -95,16 +95,16 @@
              * @ignore
              */
             me.event.subscribe(me.event.STATE_RESUME, this._resumeCallback);
-        }
+        };
 
-        init.call(this, object);
+        this.setProperties(object);
 
         /**
          * reset the tween object to default value
          * @ignore
          */
         this.onResetEvent = function ( object ) {
-            init.call(this, object);
+            this.setProperties(object);
         };
 
         /**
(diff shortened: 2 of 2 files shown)


.jshintrc
```diff
@@ -111,6 +111,9 @@
         "afterEach"             : false,
         "spyOn"                 : false,
         "xdescribe"             : false,
-        "xit"                   : false
+        "xit"                   : false,
+
+        // objects
+        "Map"                   : false
```
parasyte: You have the wrong npm dependency versions for 2.1.x, if you are hitting this issue.

agmcleod: ah okay, thought that was odd.

src/vendors/tween.js
```diff
@@ -37,52 +37,84 @@
...
+            _easingFunction = me.Tween.Easing.Linear.None;
+            _interpolationFunction = me.Tween.Interpolation.Linear;
+            _chainedTweens = [];
             _onStartCallback = null;
             _onStartCallbackFired = false;
             _onUpdateCallback = null;
             _onCompleteCallback = null;
+            _tweenTimeTracker = me.timer.lastUpdate;
+
+            _resumeCallback = function (elapsed) {
```
parasyte: I would make this an instance method instead of a function assigned to a variable via closure.

src/vendors/tween.js
```diff
@@ -37,52 +37,82 @@
...
+            }
+
+            /**
+             * Calculate delta to resume the tween
+             * @ignore
+             */
+            me.event.subscribe(me.event.STATE_RESUME, this._resumeCallback);
+        }
+
+        init.call(this, object);
```
parasyte: Is there a specific reason for using the call method, vs just `init(object)`? Same question for `onResetEvent` below.

agmcleod: Since im calling this._resumeCallback in it, needs the context of the tween, no?

parasyte: Of course! If `init` was an instance method, the bind would be unnecessary. Maybe that should be changed?

agmcleod: Probably is the direction to go. Could do that. Though I think i might rename it, since the function itself is the constructor :)

agmcleod:
@parasyte updated.

parasyte:
Ok, I like what I see. @obiot, anything to add?

obiot:
yes just one thing, as i'm confused on either we should actually use `onResetEvent` or `onActivateEvent`, since we use here `onDeactivateEvent` to unsubcribe to `STATE` related event.

agmcleod:
@obiot For the variables? nah. The reason being because we need those initialized before the start() method is invoked.

obiot:
Oh yeah indeed!

Good for me too then :)

obiot:
this one should go both to 2.1.x and 3.x however, right ?

obiot:
@agmcleod this branch can die now, right ?
https://github.com/melonjs/melonJS/tree/proposed-tween-cleanup

obiot:
@agmcleod ?

agmcleod:
@obiot i deleted it yesterday :). Typically you can safely click delete branch after merging a pull request if it asks you.

obiot:
oh yes indeed sorry, missed that as it was not yet reflected locally on my machine :P

PR #746, opened 2015-09-16
Moved pointer events into onActivateEvent and onDeactivateEvent

thedrumchannell:


agmcleod:
Awesome, thanks! Find it works well for you I hope?

Realized we don't have an example using GUI_Object in the repo. May need to update that as a test case.

PR #787, opened 2016-04-06
Oops, also update documentation.

thedrumchannell:


agmcleod:
@thedrumchannell i'm confused why this would be warranted. The function expects an object that has the method getBounds(). That is defined in the me.Rect class. Really it just returns the rect itself, but other objects will return the rect data depending on structure: https://github.com/melonjs/melonJS/blob/master/src/shapes/rectangle.js#L86

thedrumchannell:
The parameter is no longer being passed using getBounds(), but instead the item object. The change is intended to be consistent with the other quadTree method documentation such as insert() and retrieve().

agmcleod:
@thedrumchannell ah i see. I appreciate the consistency, but I feel like documenting it as any Object is simply incorrect. As passing {}, would validate that documentation, but would cause the method to fail. Same with insert, and possibly retrieve.

I've missed following a lot of the details on here lately, so if you changed those in an approved pull request, I'm not super bothered by it. I just want to ensure we keep things reasonably accurate, since with javascript it's so easy to pass invalid arguments.

parasyte:
Yeah, I think `me.Rect` is correct...

PR #792, opened 2016-05-22
Ticket 279

agmcleod:


agmcleod:
Just opening a PR so we can discuss in case you guys think there's anything missing :)

examples/font_test/main.js
```diff
@@ -62,40 +71,45 @@ var FontTest = me.Renderable.extend ({
         // arial font
         this.font = new me.Font("Arial", 8, this.color);
         // bitmap font
-        this.bFont = new me.BitmapFont("atascii", {x:8});
+        this.bFont = new me.BitmapFont("xolo12");
+        this.bFont.name = 'xolo';
```
obiot, line 75: what was this for?  is it a property of `me.BitmapFont` ?

src/font/bitmapfont.js
```diff
@@ -91,32 +95,49 @@
...
+
+        /**
+         * Measures a single line of text, does not account for \n
+         * @name measureTextLineWidth
+         * @memberOf me.BitmapFont
+         * @function
+         * @param {String} text
+         * @returns {Number} the calculated width
+         */
+        measureTextWidth : function(text) {
```
obiot, line 99: not sure we need this one as a public function, maybe I'd personally put it as a global (utility) function under `me.BitmapFont` and then pass the font data as extra argument.

src/font/bitmapfont.js
```diff
@@ -5,69 +5,73 @@
...
+            this.fontImage = me.loader.getImage(fontName);
+
+            /**
+             * The instance of me.BitmapFontData
+             * @type {me.BitmapFontData}
+             * @name bitmapFontData
+             * @memberOf me.BitmapFont
+             */
+            this.bitmapFontData = new me.BitmapFontData();
+            this.bitmapFontData.parse(fontData);
```
obiot: I was wondering if we should not do that upon loading actually, like we do with XML and then directly return a `me.BitmapFontData` object when getting back the fontData from the loader ? (a JSON would be event better, to keep the Loader data agnostic, but it's then too much level of transformation)

obiot:
added a couple of comments, but globally that's amazing, can't wait to actually play with it :)

examples/font_test/main.js
agmcleod, line 75: Was using it for debugging actually, i can delete.

src/font/bitmapfont.js
agmcleod: I'm not sure what you mean on this one. Are you thinking something like this in terms of a public facing API?

```
var data = new me.BitmapFontData(me.loader.getBinary(fontName));
data.parse(); // or whenever
var font = new me.BitmapFont(data, me.loader.getImage(fontName));
```

Or do you mean something else?

Personally I don't think we should convert to JSON. For the tile loader it makes sense, as the editor supports JSON & XML, we're just loading one as a dataset instead of maintaining two feature equal parsers.

src/font/bitmapfont.js
parasyte, line 99: `measureText` already exists for this purpose. ;)

obiot:
yeaahh!!!!

obiot:
@agmcleod would you mind modifying the platformer ? it's not working anymore since the last merge, and I"m kind of relying on it for my testing :)

obiot:
@agmcleod I updated the platformer and fixed one bug at the same time. However we need a nicer font to include here and also later in the tutorial. 

Any suggestions ? (@agmcleod @parasyte), keeping in mind that this will then probably become the "default" bitmap font of melonJS :)

obiot:
I personnally could go for that one :
http://www.dafont.com/fipps.font

and it's listed as 100% free

parasyte:
I've always liked https://www.google.com/fonts/specimen/Press+Start+2P

obiot:
yeessss, I love retro font as well, almost like the one we had before :P

agmcleod:
Hmm. i think the press start 2p might have more wide usage, but i think the fipps suits the tutorial better. I use Munro Small quite a bit with more pixel art focused games. Free use as well: https://www.fontsquirrel.com/fonts/Munro

parasyte:
Munro looks like a good VWF. And Press-Start-P2 is a good FWF. :)

obiot:
noe sure i know what VWF and FWF are  :P

agmcleod:
Variable width font & fixed width font... i think

parasyte:
Yes.

Issue #794, opened 2016-06-14
Issues with touch enabled devices with trackpads


BrianRosamilia:
Melon.js gives precedence to touch input over traditional input when determining what events to setup.

https://github.com/melonjs/melonJS/blob/037f944e59f7e387115348bea541d45275df5483/src/input/pointer.js#L216-L221

System : surface pro 3
Browser : chrome (which can't do pointer events https://bugs.chromium.org/p/chromium/issues/detail?id=471824) also see here for other browsers where this API isn't usable http://caniuse.com/#feat=pointer

This kills the mouse (trackpad) input and only allows for touch events (In the future, Apple will be releasing touch laptops so best to stay ahead of the curve (or at least give the developer the option of favoring the trackpad)

My best solution was to do the following (inside input/pointer.js)

```
 else if (me.device.touch && !me.device.isMobile) {  //  `touch****` events for touch laptops
                 activeEventList = touchEventList.concat(mouseEventList);
             }else if(me.device.touch){ //  `touch****` events for iOS/Android devices
                 activeEventList = touchEventList;
             }
```

The above code causes mousedown and touchstart to fire at the same time when using touch, so obviously it's not a workable solution.  This is due to melon being forced to simulate the pointerdown event, I'm sure, but I wasn't able to pinpoint where it's doing this.

My poor man's solution would be to check `me.device.isMobile` and then just fall into the final else statement for the legacy click events.  I'm not sure how important having dual touch/mouse events is to this project, or if anyone has even considered this!  Personally, I believe if mobile is detected as false, people won't want to be clicking their screen with their finger, but if it's worth supporting both and the simulation of pointerdown is simple to fix, my opinion may change.

I could use some help resolving this issue.

An example of this in action :

Use this example game on a surface pro 3.  Only touch will work.

http://melonjs.github.io/melonJS/examples/whack-a-mole/

Last edit :

Here's a simple way to fix this

```
 if (navigator.pointerEnabled) {
                activeEventList = pointerEventList;
            }
            else if (navigator.msPointerEnabled) { // check for backward compatibility with the 'MS' prefix
                activeEventList = MSPointerEventList;
            }
            else if (me.device.touch && me.device.isMobile) { //  `touch****` events for iOS/Android devices
                activeEventList = touchEventList;
            }
            else { // Regular Mouse events
                activeEventList = mouseEventList;
            }
```

agmcleod:
[...]

PR #795, opened 2016-06-15
[#794] Fix handling of mouse events for touch devices with mice/touch…

BrianRosamilia:
As discussed in #794


agmcleod:
Merged, and added you to authors list, as well as bug fix in change log. Thanks again!

PR #799, opened 2016-07-28
Fix for getting parent node dimensions

krojew:
As disussed in https://groups.google.com/d/topic/melonjs/sG8sGm5-E8o/discussion

### DIFF
--- a/src/video/video.js
+++ b/src/video/video.js
@@ -368,10 +368,15 @@
                 var parentNodeHeight;
                 var parentNode = me.video.renderer.getScreenCanvas().parentNode;
                 if (typeof (parentNode) !== "undefined") {
-                    // for cased where DOM is not implemented and so parentNode (e.g. Ejecta)
-                    var rect = parentNode.getBoundingClientRect();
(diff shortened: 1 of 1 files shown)


obiot:
Interesting, thanks for this ! 

However for wrapper like Ejecta that do not support `getBoundingClientRect`, and to avoid anyway a potential regression, I would test if the function exist first and then compute the value :
````javascript
if (typeof parentNode.getBoundingClientRect === 'function') {
     var rect = parentNode.getBoundingClientRect();
     parentNodeWidth = rect.width || (rect.right - rect.left);
     parentNodeHeight = rect.height || (rect.bottom - rect.top);
} else {
    // for wrapper like Ejecta not supporting getBoundingClientRect()
    parentNodeWidth = parentNode.width;
    parentNodeHeight = parentNode.height;
}
````
this said, that was a long time ago, maybe it has been "added" since then, I haven't checked

obiot:
@parasyte any comments on your side ? looks great to me !

obiot:
ok, merging this one, thanks again @krojew :)

Issue #580, opened 2014-10-09
Evaluate replacement for me.Entity.bounds


parasyte:
This seems kind of tricky. There has been a lot of discussion about this in the past, mostly in #560. But it seems like maintaining the bounds rectangle separately from the entity rectangle is problematic. See this commit as an example of how it goes wrong: https://github.com/melonjs/melonJS/commit/104b6f45a7c366a662582a796a922fde2ec51ad0 The entity geometry was silently working for most objects, but fails for TMX Polygon and Poly Line objects. :confused:

@obiot What are the reasons for creating a `bounds` object inside the `Entity`, instead of using the entity geometry itself as the bounding box?

I know the entity geometry was once used as the renderable geometry. Then that changed when renderable was broken into its own object. And then the collision box object was removed, but it appears the bounds object has just taken its place.

agmcleod:
Given how we've changed the way bounds works a bit, is this still relevant?

parasyte:
Oh yes. Something still needs to be done. :) The spec may have to evolve with recent changes, but for sure we need something better than shoving more rectangles onto entities to get everything working...

parasyte:
For future reference:

Entity:

* `getBounds()`: return a rectangle that contains all collision shapes, *or* `this` when there is no shape
* `renderable`: a rectangle that contains the visible part of the entity
* `this`: a rectangle that is a union of the above

Sprite:
* `getBounds()`: returns `this`
* `this`: a rectangle that contains the visible part of the sprite

obiot:
I was looking at remove `_bounds` from the entity object, and directly use the entity object geometry for the bounds object, but due to the following lines, i'm not sure we can :
https://github.com/melonjs/melonJS/blob/master/src/entity/entity.js#L257-L263
https://github.com/melonjs/melonJS/blob/master/src/entity/entity.js#L241-L247

as `_bounds` also serves as a "cache" where all final value (like initial position + body relative position + movements + ancestor absolute position) is included.

so as far as it goes for this ticket, what I would "only" do so far is to change the following line
https://github.com/melonjs/melonJS/blob/master/src/entity/entity.js#L263
to :
````
this._bounds.resize(w, h);
this.resize(w,h).union(this.renderable);
````

obiot:
also, are these two functions not somehow redundant :
1. `UpdateBoundsPos` : https://github.com/melonjs/melonJS/blob/master/src/entity/entity.js#L257-L263
[...]

PR #806, opened 2016-08-19
initial typescript typings

krojew:
These are (incomplete) typings for melon. Note: those were based on the source, since the official docs are very outdated.


parasyte:
FWIW, the docs should be up to date. http://melonjs.github.io/melonJS/docs/ If they are missing something, or inaccurate, please let us know what to look for, so we can fix it.

krojew:
I can't tell exactly - I've been stumbling upon inconsistencies while working on this for few days. It would be best to just go through the public api and check everything.

agmcleod:
The docs should be pretty close to master code. We have pretty minimal changes at present from 3.1. As parasyte/Jason said, please let us know if you find faults.

parasyte:
And a suggestion for this PR: If it's possible to put the type definitions inline with the code, that would be better. It's going to be impossible for us to maintain this if it has to live in a separate file. (It would be the ultimate best if TypeScript could just automate type detection, e.g. with the documentation comments.)

krojew:
Inline, as in the .js files? If so, that's not how typings work - they should be in a separate .d.ts file for existing .js code. Also, I tried syncing it with the docs, rather than the code (as it it advised to do), but that lead to inconsistencies with the code itself. And since typings should reflect the code, well...

parasyte:
In that case, I guess there's nothing to be done about the maintenance burden. Just going to expect these types to get out of date as the code evolves. The docs on the other hand, are kept in sync with the code, and are auto-generated from the doc comments (and currently synched with v3.1.0: https://github.com/melonjs/melonJS/commits/gh-pages)

You haven't provided any examples of where docs do not match code, though. If you'll point us in the right direction, we can get the inconsistencies solved.

krojew:
Well, making these took few days of work 8h/day. I don't think I will be able to go through every file again to check it all out. I think every time someone works on a public function, he should check if the comments are in sync (getting a notification to update the typings would also be great).

parasyte:
Just a couple of examples is all I'm asking?

krojew:
I'll see what I can do on Monday.

krojew:
Ok, quick examples to verify:

- Renderer.resize()
- Renderer.setAntiAlias()
- Renderer.setImageSmoothing()
- Sprite.scale() and scaleV() - wrong return type, since overriding Polygon
- Body.addShape()
- Body.getShape() - remember to always check return types
- Body.removeShape()
- Viewport.resize()

Probably more, but no time to check atm.

obiot:
did a quick pass this morning, and fixed all the one i could find :
https://github.com/melonjs/melonJS/commit/997a13c4362e37c1aa2b5b27de7b6a568d235125

not sure though what you mean with scale and scaleV, and did not find anything on addShape/removeShape

krojew:
addShape has me.Rect as an acceptable parameter type, yet getShape doesn't include Rect as a return type.

me.Sprite extends me.Rect, which extends me.Polygon. Polygon has scale methods which return itself, yet Sprite overrides those methods and don't return anything. If some code expects a Polygon as a parameter, it can assume that scale does indeed return something, but if you pass a Sprite (which is legal), you no longer maintain that contract. This would be very hard to debug in JavaScript, yet in TypeScript the compiler caught that instantly.

obiot:
for `addShape`, and `getShape`, that's because any `me.Rect` object passed to `addShape` is first converted to a proper `me.Polygon` object, so you can never get a `me.Rect` back when calling `getShape`.

very good point though for `me.Sprite`, I'll fix those too.

krojew:
Ok, I'll update the typings for that. Note that there most definitely are other such problems with the docs, but I have no time to go through it again. It would be a good idea to verify them while working on a public method.

krojew:
Just stumbled upon another thing: settings in Entity constructor don't list width or height, but those are required.

krojew:
Another one: me.loader.preload - onload callback is mentioned twice in the docs, one time is mandatory, the other is optional. In fact, it's always mandatory. Also, there is no onerror parameter.

agmcleod:
Thanks for the update there. Still does make me worried on keeping the two in sync with each other. Type file & code/api changes.

krojew:
That is a legitimate concern. Nevertheless, better to have something, even temporarily out of date, than nothing.

agmcleod:
@krojew updated the preload docs, and the entity settings docs on master. Thanks!

obiot:
I agree that something out to date is for sure better than nothing, but people will at some point complain about it being outdated, and if you are then not around none of us is (i think) using typescript and will be able to keep it up to date.

this said, it's just one big file, that will be only visible to people directly using the master branch, and won't be i guess advertised through our blog/web page ?

agmcleod:
Given we're a small group, we could convert to using it (typescript as the source) if we thought it would benefit the overall development. But yeah i think it would be best to keep it in the repo, not advertise it too much if we're not able to update it.

Could always try to manually as we document the change log.

krojew:
I, personally, would suggest moving to TypeScript. Nowadays there really is no reason at all to use "plain" JS anymore, and the benefits of moving to TS are enormous (type-safety, compile time error checking, native module and class support etc.). Also, the effort of such migration is not very big (especially given you already have most of the class interfaces done :) . In the simplest case - you can create classes and just copy-paste most of the existing code there. That being said, the decision is ultimately yours, but I haven't met anyone who switched to TS and regrets the decision.

Now, going back to the typings, you don't need to advertise them in anyway, since they are just a nice addition, not a requirement of any way. The file currently is incomplete; unfortunately, my budget isn't big enough to cover the whole library. Which is a shame, because I wanted to publish it to DefinitelyTyped.

krojew:
Next one: viewport.localToWorld() is said to take screen coordinates, while in reality it takes client coords.

agmcleod:
@krojew actually that's correct. Screen coordinates are what you consider for 0,0 being the top of the viewport and 800,600 being the bottom right (or whatever your viewport size is). Regardless of where the viewport translation is. So if you're playing the platformer tutorial, and moved over 400 pixels with the character, 0,0 will still be the top left screen coordinate.

krojew:
@agmcleod Actually, not at all. Real-life example: 0,0 coordinates in fullscreen melon for me are now approx. 2500,0. Why? Because I have melon launched on a second monitor.
Screen coordinates are something different than melon coordinates (which in turn are different than client/viewport coordinates). Testing confirms melon expects canvas coordinates, which are client coords adjusted by canvas position within the viewport. While ensuring that passed coordinates will have that adjustment is the user responsibility, the docs are in fact incorrect calling them screen coordinates.

agmcleod:
@krojew yeah i know it doesn't match the real screen, but it's a term we bring over from traditional game & graphics development :)

krojew:
Yeah, well that already caused one bug for me, and other people can fall for it too. I think it's best to stick with standardized terminology.

krojew:
Next one: Entity.pos is vector3d not vector2d, as stated in the docs.

obiot:
indeed for pos, not only we overrided the pos property from me.Rect (that is initially a Vector2d) but we indeed never declare it in the documentation. I will fix that.

to merge your PR, I will wait for you to tell that you think you are finished with it, i guess it's more simple that way ?

obiot:
wait a minute, `pos` is more like a private property and is not actually documented on purpose, @parasyte @agmcleod am I right ?

krojew:
Entity inherits Polygon, which exposes pos publicly.

krojew:
BTW. That kind of inconsistencies is another argument to switch to Typescript :)

obiot:
indeed ! :P

parasyte:
There has been a lot of discussion around how `Entity` relates to its geometry and renderable, e.g. #580, the [ECS Proposal](https://gist.github.com/agmcleod/3e901d45cd97335aa3d0), and an [experiment]() that I started, but quickly hit some road blocks.

In some ways, making `Entity` a stateless container makes sense. But the utility of the class inheriting the geometry of its children is also appealing; Our current implementation is closer to the latter, in any case.

The `Body` position is the canonical way to move an entity around. Combine that with the boundary cache, and you can imagine why the position tracking code gets really hard to follow.

obiot:
yeah, I would love to see these improvements on Entity, that would simplify so many things !

obiot:
https://github.com/melonjs/melonJS/commit/ca6ee526827ba204dd1a45f01a271e7a80c45593

@krojew btw, `pos` type is `me.ObservableVector3d` and not `me.Vector3d`

obiot:
so 4.0 is soon ready to be released, what do we do with this one, add it to the current master, or still wait a bit ?

obiot:
interesting, i was looking if there were any grunt task to automate the validation of a TS defintion file, and found this one, I just had a quick 1 minute check, but it seems to actually automatically generate it :
https://github.com/DefinitelyTyped/grunt-tsd

krojew:
It looks like something running tsd, which is now deprecated. Typescript got its own official repository of typings and they are migrating DefinitelyTyped to there (which used tsd to install typings to projects). I need to finally find the time to finish melon typings and publish here/there.

krojew:
I'm thinking of rewriting WHOLE melonJS to TypeScript, because of its current problems (baked polyfills, no npm package, strange inheritance mechanism, inconsistencies with method overrides, no type safety, etc.). The question now becomes - would you guys be interested in using it? Since TS seamlessly transpiles to any JS flavor, the end user would be able to use both the TS and transpiled JS versions.

obiot:
not sure what @parasyte and @agmcleod are thinking about it, but personally i'm really not sure about it to be honest, as the direction we talked about was rather to go for a full es6 framework in the future. Also, taking phaser as an example, they started as a typescript only framework and then switched to javascript + a TS typing definition.

krojew:
Well, phaser never was written in TS - they only supply typings (since v2, IRC). Nevertheless, I think limiting yourself to only one flavor of JS could be a potential blocker for projects. Also, you loose all the benefits of compile-time error checking and type safety (not to mention other features, but those two are the crucial ones). That and the problems with Object.extend are the main reason I'll start working on the TS version. If you're not interested in it, no problem - I'll try to keep it in sync with melon releases, if time allows.

agmcleod:
Phaser was originally written in TS, but they switched to the typings file due to having many problems with the compiler earlier on. Really these days that's less of an issue, things have gotten way more solid.

Admittedly i forgot what the issues are with Object.extend :(, but I feel like maybe addressing that functionality is maybe a solution? I just feel like you're taking on a fairly large task for one person, and I'd hate to see you get burned out by it.

krojew:
I already outlined the problems with Object here: https://github.com/melonjs/melonJS/issues/821 
I found all code interacting with melon needs more and more hacks to work. That adds a lot of development and debugging time. And the lack of actual types is also a problem. Simple analysis of the costs related to creating a more complex project in melon suggest it's actually better to rewrite everything in TS - this will be a giant boost to productivity + I bet a keg of beer the compiler will find many errors along the way. I also don't think it will be much effort - substituting Object.extends for proper classes will take the majority of work, while the logic can be mostly copy-pasted, since TS supersedes JS. The project in question involves a real-time client-server application with lots of features missing from melon ATM (e.g. post-processing), so it's way beyond a simple platformer, where the costs of custom solutions and using JS becomes irrelevant.

All in all, this will be good for melon either way. The TS version will be there for everyone using TS or any JS flavor unsupported by official builds, while the JS version will get bug fixes as a bonus :)

PR #820, opened 2016-10-08
Fix jsdoc for `me.Entity` so that `renderable` reflects that it is a me.Sprite

emarcotte:
It is slightly confusing when reading the various code/samples and comparing what APIs are available on the doc site when trying to figure out how to set up animations. My confusion was corrected when I read the source to realize `renderable` was actually a sprite and that the docs listed it as a `Renderable`.

### DIFF
--- a/src/entity/entity.js
+++ b/src/entity/entity.js
@@ -37,7 +37,7 @@
             /**
              * The entity renderable object (if defined)
              * @public
-             * @type me.Renderable
+             * @type me.Sprite
              * @name renderable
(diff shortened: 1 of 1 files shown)


obiot:
actually the original doc is more correct, as the renderable component of an entity can be any object derivating from `me.Renderable`, which includes me.Sprite or even another entity and any other user custom object derivating themselves from `me.Renderable` as well. I would rather mention that in the example rather than changing the object type in the doc.

Issue #829, opened 2016-11-09
Remove IE9/IE10 support and other legacy browser


obiot:
That includes :
- [x] deprecated / non-standard mousewheel events
https://developer.mozilla.org/en-US/docs/Web/Events/mousewheel
- [x] support for MS prefixed PointerEvent : https://github.com/melonjs/melonJS/blob/master/src/input/pointer.js#L121-L130
- [x]  Array.forEach polyfill :
https://github.com/melonjs/melonJS/blob/master/src/lang/array.js#L29-L40
http://caniuse.com/#search=forEach
- [x]  Array.isArray polyfill :
https://github.com/melonjs/melonJS/blob/master/src/lang/array.js#L42-L53
http://caniuse.com/#search=isArray
- [x] ~~TypedArray polyfill~~ : (ES6)
https://github.com/melonjs/melonJS/blob/master/src/lang/array.js#L98-L146
http://caniuse.com/#search=TypedArray
- [x] Function bind polyfill : (partial support in older version of android < 4.3)
https://github.com/melonjs/melonJS/blob/master/src/lang/function.js#L14-L59
http://caniuse.com/#search=Function.prototype.bind
- [x] Object defineProperty polyfill :
https://github.com/melonjs/melonJS/blob/master/src/lang/object.js#L15-L38
http://kangax.github.io/compat-table/es5/#test-Object.defineProperties
- [x] Object Create polyfill :
https://github.com/melonjs/melonJS/blob/master/src/lang/object.js#L40-L62
http://kangax.github.io/compat-table/es5/#test-Object.defineProperties
- [x] ~~Object is Polyfill~~ (ES6) :
https://github.com/melonjs/melonJS/blob/master/src/lang/object.js#L64-L102
http://kangax.github.io/compat-table/es6/#test-Object_static_methods_Object.is
- [x] ~~Object.assign~~ (ES6) :
https://github.com/melonjs/melonJS/blob/master/src/lang/object.js#L104-L152
http://kangax.github.io/compat-table/es6/#test-Object_static_methods_Object.assign
- [x] ~~Date Performance polyfill~~ : (not supported by opera mini, and android < 4.2, might need to be extended to support webworkers)
https://github.com/melonjs/melonJS/blob/master/src/lang/performance.js#L23-L39
http://caniuse.com/#search=performance
https://developer.mozilla.org/en-US/docs/Web/API/Performance/now
- [x] Date now :
[...]

PR #842, opened 2016-11-30
Pointer event fixes

agmcleod:
Tested against shapes with desktop & mobile touch simulation. Tested the multitouch with browser & iOS simulator as well.


obiot:
did you also tested on window based devices (that rely on PointerEvent) ?

I guess anyway you can merge into master ? 4.0.0 is out and I already bumped the version to 4.0.1

once merged we can provide a [WIP build](http://melonjs-builds.s3.amazonaws.com/index.html?prefix=artifacts/), and see if it also fixes David's [https://groups.google.com/forum/#!topic/melonjs/Gx6sH1SN3BI](issue on the forum).

obiot:
my god i hate that systematically avec every release, we have a major bug fix coming, ahahahah :)

parasyte:
To be fair, we also have no test coverage for pointer events.

obiot:
I actually thought about it a few weeks ago, but I'm not sure how to do it, although i guess we can some hackish code in the test unit that fake some pointer move ??

parasyte:
Mocks. Events can be created with the [CustomEvents API](https://developer.mozilla.org/en-US/docs/Web/API/Event/Event) The mock can send these events to trigger any event listener.

parasyte:
There's something of a tutorial on custom events here: https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Creating_and_triggering_events

agmcleod:
Not a bad idea. Also I don't really have a windows device to test it on for touch events, where the bug originally came from. I can run it on chrome though and check.

obiot:
so we can merge this right ? or you want to add some basic pointerEvent test unit together with the PR ?

obiot:
don't forget the CHANGELOG too :P

obiot:
hi guys,

btw, didn't we agree we could remove support for the prefixed MS pointer event ?
https://github.com/melonjs/melonJS/blob/master/src/input/pointer.js#L130-L138
https://github.com/melonjs/melonJS/blob/master/src/input/pointer.js#L221-L222

obiot:
yes / no ?

obiot:
also listed here : #829

Issue #844, opened 2016-12-05
me.levelDirector.loadLevel fires callback two times


krojew:
Hi,
I've encountered a strange nondeterministic behavior. Sometimes when refreshing the page with melon running, calling me.levelDirector.loadLevel _once_ causes the onLoaded callback to fire (at least) _twice_. I really don't know where it's coming from and why, since the stack comes from some async callback which calls safeLoadLevel.

krojew:
I think I'm getting closer to tracking it down. It might be related to me.loader.preload() - it calls checkLoadStatus() which schedules itself to be called in the future and then call global api.onload, when everything is loaded. There might be some bad global state/timing problem when me.loader.preload() gets called two times with different callbacks.

obiot:
maybe we should clean it up when the page is unloaded
https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload

not sure though it works in case of refresh

krojew:
This won't help. Seems like the checkLoadStatus() function isn't reentrant because of the global status, which implies me.loader.preload() isn't reentrant. I think this is the case I've stumbled upon, and it would explain why it's nondeterministic.

krojew:
Did a quick fix - seems to work now.

krojew:
I'm happy my PR was merged, but be aware it's not the final fix - just a quick workaround. To make me.loader.preload really reentrant, it should never touch any global state. The same goes for any function which takes a callback.

Well, since I'm already talking about it, I might as well make a proper fix for this one :)

PR #846, opened 2016-12-05
Made loader reentrant

krojew:
ticket #844

### DIFF
--- a/src/loader/loader.js
+++ b/src/loader/loader.js
@@ -42,8 +42,10 @@
                     // make sure we clear the timer
                     clearTimeout(timerId);
                     // trigger the onload callback
+
+                    var onload = api.onload;
                     setTimeout(function () {
-                        api.onload();
+                        onload();
                         me.event.publish(me.event.LOADER_COMPLETE);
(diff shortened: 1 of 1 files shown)


obiot:
I'm not sure to get how it fixes it, but happy to merge it if it does :) thanks for the PR!

parasyte:
I'm confused by it, too. Does the bug occur as a result of dereferencing the `onload` method after the timeout fires? And caching the method at timer-creation resolves some race condition? Even then, still not clear how this prevents it from a double-call.

krojew:
This doesn't prevent a double call, rather calls the proper callback. Consider the situation:

1. Call preload and pass a callback.
1. Call preload again and pass other callback.
1. Both loads finish and the second callback is called twice.

The reason is mixing async callbacks with the lack of reentrancy - this sooner or later had to result in an error. I suggest looking at other places where some global state is being used along with async calls.

parasyte:
Oh, you are calling `loader.preload()` more than once. Understood. These changes for #844 are :+1:

PR #847, opened 2016-12-05
Proper handling of local onload callback

krojew:


krojew:
ticket #844

obiot:
:+1 :)

src/loader/loader.js
```diff
@@ -35,17 +35,17 @@
...
                 // wait 1/2s and execute callback (cheap workaround to ensure everything is loaded)
-                if (api.onload) {
+                if (onload || api.onload) {
                     // make sure we clear the timer
                     clearTimeout(timerId);
                     // trigger the onload callback
-
-                    var onload = api.onload;
+                    // we call either the supplied callback (which takes precedence) or the global one
+                    var callback = onload || api.onload;
```
parasyte, line 46: This can be hoisted to the top of function scope, so you don't have to falsy-check `onload` twice.

obiot: i would even say that since checkLoadStatus is always called passing onload, there is no real need for that

parasyte: Excellent observation! The function is private (undocumented), so none of the fallback logic is even necessary.

obiot: thanks, I still have my moments sometimes :P:P:P

Issue #872, opened 2017-02-12
bindPointer touchstart doesn't fire


Ld00d:
When the code gets to line 585 of pointer.js:

`if (e.type === POINTER_DOWN[0] || e.type === POINTER_DOWN[1] || e.type === POINTER_DOWN[2]) `

It's missing the check for touchstart:

`var POINTER_DOWN = ["pointerdown", "MSPointerDown", "mousedown", "touchstart"];`

I can fix it and submit a pull request.

obiot:
good catch and thank you for the pull request :)

PR #873, opened 2017-02-12
[#872] Add touchstart to pointer event check

Ld00d:
Just adding the 4th element of the POINTER_DOWN array to the check for bound pointer events.

### DIFF
--- a/src/input/pointer.js
+++ b/src/input/pointer.js
@@ -582,7 +582,7 @@
 
         // check if mapped to a key
         if (keycode) {
-            if (e.type === POINTER_DOWN[0] || e.type === POINTER_DOWN[1] || e.type === POINTER_DOWN[2] || e.type === POINTER_DOWN[3]) {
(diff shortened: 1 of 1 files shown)


obiot:
good catch ! i cannot even believe nobody saw this before, I guess this is another regression following the small refactoring in 4.0.0.

maybe we can shorten the code though by using a 
```` javascript
if (POINTER_DOWN.indexOf(e.type) === -1) {
   ....
}
```

thank you very much, if you could just do this last change, and then we can merge it :)

parasyte:
```javascript
if (POINTER_DOWN.indexof(e.type) >= 0) { /* ... */ }
```

But yeah, for sure, I also think this patch is needed.

Ld00d:
I considered indexOf, but I thought maybe it was a performance concern. I'll make the change...

parasyte:
`indexOf` will probably be faster, because the iteration can be done in native code. Checking each index of an array in JavaScript is just an unrolled iteration loop. ;)

Issue #810, opened 2016-09-06
ES6 Collections tests are failing under Jasmine 


obiot:
this only happen when running the test in the browser (by executing  the test [index.html](https://github.com/melonjs/melonJS/blob/master/tests/index.html) file).
- **ES6 Collections test Map#set** : 
  - _Expected true to equal false._
  - _Expected Function to be undefined._
  - _Expected Object({  }) not to equal Object({  })._
- **ES6 Collections test Map#forEach with mutations** :
  - _Expected undefined to equal '[1,3,0]'._

the reason being that the es6 collections is deprecated and not fully aligned anymore with the standard : https://github.com/WebReflection/es6-collections 

> This  is a very old polyfill which served old browser decently for the last 4+ years. I am not actively maintaining this project but I would eventually accept PRs if you really need to use it.
> 
> Please have a look at better alternatives such [ES-Shims](https://github.com/paulmillr/es6-shim)

Solution(s) :
- push a PR, a modify our local version ?
- change for a more up-to-date shim ?
- dont'do anything, this is not impacting melonJS itself, and at some point we should be able to remove the shim ?

parasyte:
Solution: Remove the polyfill in #829 :) `Map` is now supported in all the browsers we target!

obiot:
as much as I agree with you on the rest (#829), i'm not sure this is yet ready to be removed for mobile browser, and especially Android 5.0 and lower :
http://kangax.github.io/compat-table/es6/#test-Map

not a problem on IOS though, as most of the installed based has migrated already

PR #890, opened 2017-07-01
Added test execution on Chrome browser. Updated grunt, grunt plugins. Fixed new ESLint warning.

kibertoad:


kibertoad:
3 tests are failing when executed on Chrome, but I'm not sure if that's not due to differences between PhantomJS and Chrome. Those can be looked at separately.

obiot:
wow, that's great ! I was actually looking at karma the other day and was wondering if it would work as well with the travis build process (In this case Chrome is not really headless, compared to phantomJS)

obiot:
failing tests are on:
* the ES6 Map polyfill (@see https://github.com/melonjs/melonJS/issues/810), maybe we should just remove those tests, i don't know.... 
* pointerEvent, this one should work though as they are suppossed to be supported since Chrome 55

kibertoad:
@obiot Note the https://github.com/karma-runner/karma-chrome-launcher/pull/111 -> karma-chrome-launcher used in this PR already includes the 111, so making this CI-friendly should be a matter of just specifying a different launcher.

PR #891, opened 2017-07-08
Added code coverage calculation for Karma tests

kibertoad:


obiot:
👍

kibertoad:
@obiot Thanks! Any chance you could take a look at https://github.com/kibertoad/melonJS/tree/kibertoad/tests and comment on what could be happening, why I'm getting "Cannot instantiate entity of type 'me.Matrix2d'"? I imagine this is related to order of test execution being changed, but I can't pinpoint what is the exact reason :(.

obiot:
i think i had this issue before, and the problem was melonJS not being initialized properly before the running the test. me.Matrix2d relies on the object pooling, and if the pooling system is not initizialized, calling the me.Matrix constructor will fail

kibertoad:
@obiot Where does the melonJS initialization happen? I've triple checked that init.js block is executed before the test, that doesn't seem to be enough.

kibertoad:
@obiot Wait, that was actually enough information, I've finally found out what exactly was happening, fixing now. Thanks!

obiot:
I just realised that karma was failing at launching both phantoms and chrome on OSX (Sierra).

After some digging, i found out that this was due to my /etc/hosts file not being formatted correctly, as `::1` was define for localhost instead of `127.0.0.1`.

Not sure this is specific to my computer, or if anyone else had the same issue, but I will check on my other mac when i have the chance.

@parasyte @agmcleod if by any chance you can find the time to test on your side :)

agmcleod:
Haha nyan cat when i run it in node XD.

How do i run the tests in chrome? I ran grunt serve, but the localhost url is just the examples page. I forget these things :P

obiot:
ahahah yes, i love it too :P
replace `PhantomJS` by `ChromeHeadless` here :
https://github.com/melonjs/melonJS/blob/master/Gruntfile.js#L241

agmcleod:
```
Failed Tests:
 ES6 Collections test
    Map#set
       HeadlessChrome 0.0.0 (Mac OS X 10.12.5)
          1) Expected true to equal false.
             at Object.<anonymous> (tests/spec/es6-collections.js:163:22)

    Map#forEach with mutations
       HeadlessChrome 0.0.0 (Mac OS X 10.12.5)
          2) Expected undefined to equal '[1,3,0]'.
             at Object.<anonymous> (tests/spec/es6-collections.js:260:39)


     292 total   286 passed   2 failed   4 skipped

 LOG MESSAGES FOR: HeadlessChrome 0.0.0 (Mac OS X 10.12.5) INSTANCE #: 52619741
    'Initializing melonJS'

Warning: Task "karma:unit" failed. Use --force to continue.

Aborted due to warnings.
```

obiot:
yes that's normal, the es6-collections polypill is outdated, and therefore the test result are not "correct" when running in a es6 compliant browsers.

I'm about to commit a change and disable the corresponding tests

PR #893, opened 2017-07-08
Removed hardcoded test list, added new test

kibertoad:
Admittedly, this PR is somewhat opinionated, hence please let me know if you'd like to have part of it dropped.

Here is the list of changes with reasoning behind why I think they make sense:

1) Executed tests are now selected based on path, not a predefined list. Hardcoded JSON of executed tests is fragile, people will always forget to add their test to it;
2) Variable initialization that depends on MelonJS being initialized was moved into BeforeAll blocks so that their execution order can be handled using Jasmine execution hierarchy mechanism. This not only removes dependency on having hardcoded execution list, but also allows to reset those variables before each test when needed (e. g. in Colour test that actually makes sense due to state mutation);
3) grunt-jasmine task dropped, it probably makes no sense to maintain its configuration alongside Karma which is definitely more flexible and powerful;
4) Helper file moved outside of specs folder. Probably init.js should be moved out as well so that only actual tests are stored there.
5) New test (/utils/util-spec) path structure mimics the path structure of the file that it tests (/utils/util.js) - in the future it might make sense to modify paths for all tests in same fashion. It makes connecting test to what exactly it is covering much easier.


kibertoad:
@obiot Ready for review! Please let me know if you'd like me to change something.

obiot:
love the changes man, thank you so much for your awesome help on this, I must admit i'm not an expert on automating test suite, as it was mainly the work of @parasyte, but I don't see anything wrong with it on my side !

agmcleod:
I do a fair bit with mocha at work, this looks good to me :).

kibertoad:
Thanks!

Issue #917, opened 2018-01-24
Please support for wx small game.


qichunren:
I love beautiful & light weight melonJS.

Wx small game is running in embed custom js runtime. There is no DOM.  It use JavaScriptCore in iOS platform. and V8 in android.

There is only a custom canvas object for developer to render game objects.

Wx small game load game js files use CommonJS style:

Need use "module.exports" in melon js lib to export API.

Then in game.js  script:

require('./melonjs')

https://mp.weixin.qq.com/debug/wxagame/dev/index.html?t=2018115

obiot:
Hi and thank you !

If I'm correct there are 2 requests here : 
1. support the wx api ;  for example they have a proprietary create canvas method
2. support modules, this is in the plan for the 6.0.0 branch (see as well #915)

am I correct?

qichunren:
I test melonJS  v5.1.0 on WX small game platform v1.02.1801081 on MAC, currently, I failed to run a demo on it.

There are some problems currently I found in melonJS.

1. Property and functions from window object should not be used, there is no DOM in WX runtime. Such as 'scroll' and 'hash' I found currently.

2. Canvas init should invoke "wx.createCanvas()"

3.  

```
JSON.parse(localStorage.getItem("me.save")) || [];
```
failed when empty value.


3.  
```
accumulator += me.timer.getDelta();   
````
var accumulator will negative when it too big, this cause game objects not invoke update method.

After I fixed above problems, I still see no game stage created, it is just blank screen. There is no error in consle. But I think the game engine update cycleing is running OK, maybe some problem related about render.



I will later send PR later maybe.

obiot:
@qichunren hi I saw you closed the PR, are you still working on it or you given up ?

qichunren:
@obiot My PR currently based on master branch, I will later edit my PR to fork from 5.x.x after fixed game render working.

Currently I have fixed blank screen display on game stage. But there is some problem about  game rect height

obiot:
Awesome ! 

That’s weird about the game rect, did you try to disable scaling to see if maybe this is related ?

obiot:
@qichunren  any update ? I'm ok if you want to submit a partial support through a PR (on the 5.x branch), always good to have something in, and can be completed later (would be sad to lose all the effort you put into this so far)

obiot:
https://medium.com/@yelin.qiu/a-complete-manual-on-wechat-mini-program-development-8fd28a85ee0d

[...]

Issue #915, opened 2018-01-11
ES6 class definition & module support


obiot:
This ticket list all the related tasks required to migrate to es6 and using a module approach. Until fully stable, the current build process will however keep transpiling everything to es5 and generate a es5 umd module. Ultimately the goal is to generate two different build :
- `melonjs.js` legacy es5 build for legacy browsers/devices (to be decided yet at which level we keep some backward compatibility with regard to the current Jay inheritance mechanism vs a standard es5 here)
- `melons-module.js` a es6 module based version for modern browser/devices (use es6 inheritance mechanism)
 
Migration phases :

* [ ] **ES6 class semantic** : to replace the current Jay inheritance mechanism (see also #685) : 
    *  [ ]  me.Object
    *  [ ]  me.Ellipse : -> me.Object
    *  [ ]  me.Pointer (used by me.input) : -> me.Object
    *  [ ]  me.Rect : -> me.Polygon
    *  [ ]  me.Polygon : -> me.Object
    *  [ ]  me.Line : -> me.Polygon
    *  [ ]  me.Renderable : -> me.Object
    *  [ ]  me.Camera2d : -> me.Renderable
    *  [ ]  me.Sprite : -> me.Renderable
    *  [ ] me.ColorLayer : -> me.Renderable
    *  [ ] me.ImageLayer : -> me.Renderable
    *  [ ] me.GUI : -> me.Renderable
    *  [ ]  me.Body : -> me.Rect
    *  [ ]  me.Entity (deprecated in version 9.0) : -> me.Renderable
    *  [ ]  me.Text : -> me.Renderable
    *  [ ]  me.BitmapText : -> me.Renderable
    *  [ ]  me.BitmapTextData (only used by me.BitmapText) : -> me.Renderable
    *  [ ]  me.Container : -> me.Renderable
    *  [ ]  me.Stage : -> me.Renderable
* [x] ~~**ES6 module export** : to allow generating a es6 bundle: 
    *  [x]  Class definition
        *  [x] me.Object
        *  [x] me.Renderable
        *  [x] me.Sprite
        *  [x] me.ColorLayer
        *  [x] me.ImageLayer
        *  [x] me.GUI
        *  [x] me.Entity (deprecated in version 9.0)
        *  [x] me.Text
        *  [x] me.BitmapText
        *  [x] me.BitmapTextData (only used by me.BitmapText)
        *  [x] me.Container
        *  [x] me.Stage
        *  [x] me.Vector2d
[...]

PR #918, opened 2018-01-24
Wx small game adapted [WIP]

qichunren:
I try to make melonJS run on WX small game platform. This PR is not complete to fixed problem, I try to work on it.  #917

### DIFF
--- a/src/game.js
+++ b/src/game.js
@@ -281,7 +281,7 @@
                 me.input._updateGamepads();
 
                 accumulator += me.timer.getDelta();
-                if(accumulator < 0)
+                if (accumulator < 0)
                 {
                     accumulator = 0.0;
                 }
--- a/src/game.js
+++ b/src/game.js
@@ -281,6 +281,10 @@
                 me.input._updateGamepads();
 
                 accumulator += me.timer.getDelta();
+                if(accumulator < 0)
+                {
+                    accumulator = 0.0;
+                }
                 accumulator = Math.min(accumulator, accumulatorMax);
 
                 updateDelta = (me.sys.interpolation) ? me.timer.getDelta() : stepSize;
--- a/src/loader/loader.js
+++ b/src/loader/loader.js
@@ -73,13 +73,20 @@
          */
         function preloadImage(img, onload, onerror) {
             // create new Image object and add to list
-            imgList[img.name] = new Image();
+            
+            if (me.device.isWxSmallGame) {
+                imgList[img.name] = me.device.WxSmallGame.createImage();
+            } else {
+                imgList[img.name] = new Image();
+            }
+            
             imgList[img.name].onload = onload;
             imgList[img.name].onerror = onerror;
             if (typeof (api.crossOrigin) === "string") {
                 imgList[img.name].crossOrigin = api.crossOrigin;
             }
             imgList[img.name].src = img.src + api.nocache;
+            console.log("imgList[img.name].src", img.name, imgList[img.name].src);
         }
 
         /**
@@ -688,7 +695,9 @@
          * @return {HTMLImageElement}
          */
         api.getImage = function (image) {
-            if ((image instanceof HTMLImageElement) || (image instanceof HTMLCanvasElement)) {
+            // TODO: maybe a better method to compare image value type.
+            // (image instanceof HTMLImageElement) || (image instanceof HTMLCanvasElement) not work on WxSmallGame playform
+            if ( typeof(image) !== "string") {
                 // if the given parameter is already an Image object
                 return image;
             } else {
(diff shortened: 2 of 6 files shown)


qichunren:
Now game engine update loop  is run as expect, but no game object display on screen. Mybe caused by Tencent wechat game canvas draw mechanism. 

It used onscreen canvas and offscreen canvas mechanism:

```
var screenCanvas = wx.createCanvas()
var offScreenCanvas = wx.createCanvas()
var offContext = offScreenCanvas.getContext('2d')
offContext.fillStyle = 'red'
offContext.fillRect(0, 0, 100, 100)
```

In order to draw red rect on screen， need to append offScreenCanvas to real canvas:

```
var screenContext = screenCanvas.getContext('2d')
screenContext.drawImage(offScreenCanvas, 0, 0)
```

thanks.

obiot:
Hi, First comment here (more coming directly on the changes) : don't you want to make this PR on the last "stable" [5.x branch](https://github.com/melonjs/melonJS/tree/5.x.x) ? 

master is the development branch, and the current 6.0 will go under some major [changes](https://github.com/melonjs/melonJS/issues/915) that will probably break it on a regular basis :)

src/game.js
```diff
@@ -281,6 +281,10 @@
                 me.input._updateGamepads();
 
                 accumulator += me.timer.getDelta();
```
obiot, line 283: would probably be interesting though to understand why `getDelta()` returns a negative value...

src/loader/loader.js
```diff
@@ -688,7 +695,9 @@
          * @return {HTMLImageElement}
          */
         api.getImage = function (image) {
-            if ((image instanceof HTMLImageElement) || (image instanceof HTMLCanvasElement)) {
```
obiot: yeah but if you remove this, this is breaking the `getImage()` behaviour for a lots of things around the framework.... why not rather add `if (weixinSmallGame && ! String)` ?

src/system/device.js
```diff
@@ -4,6 +4,10 @@
  * http://www.melonjs.org
  *
  */
+
+/*global wx*/
+/*eslint no-undef: "error"*/
```
obiot: global should be declare here : 
https://github.com/melonjs/melonJS/blob/master/.eslintrc.json#L9

and please avoid adding more aslant exception

src/system/device.js
```diff
@@ -22,7 +26,12 @@
         var swipeEnabled = true;
         var disableSwipeFn = function (e) {
             e.preventDefault();
-            window.scroll(0, 0);
+            if (!me.device.isWxSmallGame)
+            {
+                // There is not DOM in WxSmallGame runtime.
+                window.scroll(0, 0);
+            }
+
```
obiot: since disableSwipeFn is a user called function, I don't see the need to add this check here (melonJS never calls it himself). And the test should rather be in your game to call it or not if you need this on a standard device/browser

src/system/device.js
```diff
@@ -225,6 +234,15 @@
...
                                  me.device.Kindle || false;
+
+            // Wechat small game platform
+            me.device.isWxSmallGame = (typeof wx !== "undefined");
+
+            if (me.device.isWxSmallGame)
+            {
+                me.device.WxSmallGame = wx;
+            }
+
```
obiot: we can just check `me.device.isWxSmallGame` around to see if we are running on the weixin platform, no need to create another `me.device.WxSmallGam` alias to `wx` ?

src/system/device.js
```diff
@@ -429,6 +447,24 @@
...
         api.Kindle = false;
 
+        /**
+         * equals to true if the device is running under wechat small game platform.
+         * @type Boolean
+         * @readonly
+         * @name isWxSmallGame
+         * @memberOf me.device
+         */
+        api.isWxSmallGame = false;
```
obiot: don't you prefer a more explicit name . like WeChat or WeChatSG ?

src/system/device.js
qichunren: OK, I follow, change later in PR.

src/game.js
qichunren: I am confused. But I debug it with Wechat game dev tool, from its console, I found it negative

src/system/device.js
qichunren: OK, I agree.

src/system/device.js
qichunren: OK, use WeChat

qichunren:
Wechat game platform support ES6, it auto translate ES6 to ES5 in its dev tool.

Could you help fix no object display on game stage currently on Wechat game platform. I am confused，debugg for two days, found no way to solve it.

Issue #919, opened 2018-01-30
onComplete is not triggered when a sprite animation contains only one frame


obiot:
as reported by @adrianhartanto0 on Gitter, if only one frame index is passed to the addAnimation method (2nd param), the corresponding onComplete callback is not triggered

adrianhartanto0:
@obiot I can work on it

obiot:
👍 make sure you apply for a PR against the 5.x branch, as the master (development) one is going under heavy changes :)

obiot:
note sure that adding an onstep method is really the best solution to be honest .... by the look of it, I'd rather say that the following [if statement](https://github.com/melonjs/melonJS/blob/5.x.x/src/renderable/sprite.js#L466-L480) could probably be split and moved outside of the parent [if statement](https://github.com/melonjs/melonJS/blob/5.x.x/src/renderable/sprite.js#L466-L480).

also if you do provide a Pull Request for this one, it would be great if you could also complete the corresponding test unit to test the onComplete callback for all possible cases (missing for now)
https://github.com/melonjs/melonJS/blob/5.x.x/tests/spec/sprite-spec.js

adrianhartanto0:
The [culprit](https://github.com/melonjs/melonJS/blob/5.x.x/src/renderable/sprite.js#L457) seems to be in that if statement where the `length` of `this.current` is `not greater than 1`, when the sprite array only contains one element/index

adrianhartanto0:
Mind elaborating what you mean by all possible cases? All I can think of is generate a sprite array of random length and check whether the callback is called @obiot

obiot:
all possible cases for now are limited : 
- no animation (me.Sprite with no animation defined), this is similar to a one frame animation except that this should not trigger the onComplete callback
- 1 frame animation
- multiple frame animation
- take in account potential delay (should the callback be triggered when the last frame is displayed, or at when the last frame is displayed + the specified delay if any)
:)

PR #920, opened 2018-02-09
fix onComplete of setCurrentAnimation not called

adrianhartanto0:
#919


obiot:
awesome, pretty simple fix actually ! thank you for the contribution :)

Issue #824, opened 2016-10-26
me.Sprite internal/private data clean-up


obiot:
Following the merge of me.Sprite and me.Animation, it would be nice as well to review a bit how we store the animation definition data within a me.Sprite object.

today it's working just fine, but it could greatly be optimized and written a bit better (cleaner).

typical issue, where I don't know what is happening :
https://github.com/melonjs/melonJS/blob/master/src/renderable/sprite.js#L378

adrianhartanto0:
Mind elaborating of what you have in mind? @obiot

obiot:
the way the animation data (frame index, etc...) is stored in the sprite object needs a good clean-up/rewrite. 

There are some quick hacks to make it working, see here for example :
https://github.com/melonjs/melonJS/blob/master/src/renderable/sprite.js#L412-L425

adrianhartanto0:
Looking through the src, I saw the the `this.current` is rather redundant as there is already `this.anim`. To replace the this.current, I thought of just keeping only the name of current animation frame. So any modification to `this.current` such as in `setAnimationFrame` will just modify the `this.anim[currentAnimationName]` @obiot

obiot:
yes, most probably some redundant stuff there :)

obiot:
note that `this.anim` is supposed to contains all the index/delay of a given animation (defined through setAnimation), while `this.current` only contains the current frame to be displayed

adrianhartanto0:
@obiot Sorry I submit another commit without reading you comment in here. So my commit isn't suitable?

obiot:
this was done it 5.0

PR #921, opened 2018-02-15
cleanup me.Sprite

adrianhartanto0:
Temporary solution for #824. I removed `this.current` and remove some unnecessary props that are only referenced in Sprite itself and not referenced anywhere else.

### DIFF
--- a/src/renderable/sprite.js
+++ b/src/renderable/sprite.js
@@ -61,7 +61,7 @@
              * @name offset
              * @memberOf me.Sprite
              */
-            this.offset = new me.Vector2d();
+            // this.offset = new me.Vector2d();
 
             // hold all defined animation
             this.anim = {};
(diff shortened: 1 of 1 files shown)


src/renderable/sprite.js
```diff
@@ -560,8 +584,7 @@
 
             // frame offset in the texture/atlas
             var frame_offset = frame.offset;
-            var g_offset = this.offset;
-
+            var g_offset = new me.Vector2d()
```
obiot: definitely not good here, as now me.Sprite is creating a new Vector object at each frame !

obiot: also this.offset is global offset applied to any frame, if you redefined here g_offset, then this is a feature regression.

obiot:
Hi Adrian,

Sorry for the late feedback,  can you clean all the comments, as it makes it difficult to see where are the changes ? see as well further comments/questions in the code changes.

src/renderable/sprite.js
```diff
@@ -69,17 +69,20 @@
             // a flag to reset animation
             this.resetAnim = null;
 
+            // name of current frame
+            this.currentAnimName = null;
```
obiot, line 73: should this one not set to "default" or something like this in case there is no animation defined (like just a fix sprite) ? Based on the current code, `setRegion` would certainly fail if you try to change the current region of a defined fixed sprite

obiot: if you could as well add a test case for it in the corresponding test unit, that would be awesome, just to be sure we cover all possible use case with me.Sprite and ensure no regressions.

obiot:
this one looks almost done once those last comments taken in account, thank you !

adrianhartanto0:
@obiot For the cleaning of comments, shall it include those you made or just those as a result of my changes?

obiot:
Let’s remove all commented code, or comments to highlight some issues that were adressed now :)

src/renderable/sprite.js
adrianhartanto0: I assumed by defined fixed sprite you mean a sprite object that has only one frame/no sprite-sheet. As such, I am adding a check in `setRegion`, if the `this.currentAnimName` is null, I just add a dummy animation named `default` and set `this.currentAnimName` to the default str.

obiot:
Hi, will u submit an update on this one ? :) thanks

obiot:
I guess this is a no then ?

PR #922, opened 2018-02-20
Fix bounds

### LINKED ISSUE #848: Disappearing entities
When an entity gets out of view it's not rendered. This is general is ok, but the method of determining if the entity is in view is incorrect. It either ignores bounds or ignores the anchor point. I have entities which anchor point is in the center of their images and they disappear as soon as the center gets out of view, ignoring the fact that the half of the image is still in the viewport.

adrianhartanto0:
To close #848. I am not completely sure if this is the best solution. There is one test case in `droptarget-spec.js` that fails and I'm not rather sure how to fix it. Waiting for your comments. I haven't deleted Entity in this case. I'll do it once this solution is alright

### DIFF
--- a/plugins/debug/debugPanel.js
+++ b/plugins/debug/debugPanel.js
@@ -328,6 +328,8 @@
                         var ax = this.anchorPoint.x * bounds.width,
                             ay = this.anchorPoint.y * bounds.height;
 
+                        renderer.save();
+
                         // translate back as the bounds position
                         // is already adjusted to the anchor Point
                         renderer.translate(ax, ay);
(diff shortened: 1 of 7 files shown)


obiot:
Hi Adrian, looks good ! 

Couple of things :
- I see you removed the body drawing from the debugPanel, this should now be added back in case its' existing on any renderable :)
- let's not remove me.Entity, to at least provide some backward compatibility in terms of API
- draggable should work, I will look into it, also keep in mind that [isKinematic](http://melonjs.github.io/melonJS/docs/me.Renderable.html#isKinematic) needs to be set to `false` to enable pointer event

obiot:
I also commented on the other pull request (I see this one also includes the changes on me.Sprite, so let's try to close the other one first, so that this one can reflect it too)

obiot:
see #1008

PR #923, opened 2018-04-05
Gui_object doesn't play animation

YosuaHamonangan:
My name is Yosua and I'm from Melon Gaming. 

I noticed that the gui object in melonJS can't play animation even though it extends sprite. I try adding super function to gui_object's update function and the animation can start working.

### DIFF
--- a/src/renderable/GUI.js
+++ b/src/renderable/GUI.js
@@ -110,7 +110,7 @@
          * return true if the object has been clicked
          * @ignore
          */
-        update : function () {
+        update : function (dt) {
             if (this.updated) {
                 // clear the flag
                 if (!this.released) {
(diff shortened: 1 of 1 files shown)


obiot:
great ! thanks :)

can you however resend it against the 5.x.x ? (master branch is used revamped to fully comply with ES6)

YosuaHamonangan:
Okay, I have resend the request

obiot:
here is the proper fix for it :
https://github.com/melonjs/melonJS/commit/f5444247ffb1f36a6ff61290c8dd09a278d090e2

note :
- your fix was only partially working (animation was still not updated in case the GUI object was clicked)
- see the proper syntax to call a parent method in my version (vs yours)

nevertheless, thank you for reporting it ! :)

PR #924, opened 2018-04-13
Gui_object doesn't play animation

YosuaHamonangan:
resend for gui_object doesn't play animation fix


obiot:
sorry but you are dragging along commits from the 6.0.0 branch into this one....

PR #928, opened 2018-05-28
Use me.Math.* instead of depricated/removed Number.prototype.*

ydm:
\+ few examples updated


obiot:
great, thank you very much !

PR #931, opened 2018-06-30
Fix artifact uploads on travis

parasyte:
- The nodeJS version was updated in `.travis.yml`, but not updated here


obiot:
good one indeed

PR #935, opened 2018-07-03
flip relative to the middle axis of a renderable

YosuaHamonangan:
I would like suggest a different way to flip renderable, because the current flip function doesn't work well with me.sprite since it flips relative to the x or y axis of the world. This flip function is mainly done in the predraw function and will take into account the current position and anchor point of the renderable so that the flipped sprite stays in its place


obiot:
all right, looks good indeed. How does it work with the current examples, and when scaling anything ?

PR #939, opened 2018-07-20
Floating renderable's bounds still follows its container position

YosuaHamonangan:
When a renderable is floating, the location of the image and the bound is different (the image is floating but the bound still follow its container). I found problem like gui_object become unclickable because of this. However I'm not sure if this can affect other function


obiot:
good one 👍  any chance you could add a corresponding test somewhere in the test units ?

PR #941, opened 2018-07-24
Different tile size for different tile layer

YosuaHamonangan:
I need this because I use different tile size for different tile layer in Melon Editor

### DIFF
--- a/src/level/TMXLayer.js
+++ b/src/level/TMXLayer.js
@@ -65,8 +65,8 @@
             this._super(me.Renderable, "init", [0, 0, 0, 0]);
 
             // tile width & height
-            this.tilewidth  = tilewidth || data.tilewidth;
-            this.tileheight = tileheight || data.tilewidth;
+            this.tilewidth  = data.tilewidth || tilewidth;
+            this.tileheight = data.tileheight || tileheight;
 
             // layer orientation
             this.orientation = orientation;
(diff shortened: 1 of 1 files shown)


src/level/TMXLayer.js
```diff
@@ -65,8 +65,8 @@
             this._super(me.Renderable, "init", [0, 0, 0, 0]);
 
             // tile width & height
-            this.tilewidth  = tilewidth;
-            this.tileheight = tileheight;
+            this.tilewidth  = tilewidth || data.tilewidth;
+            this.tileheight = tileheight || data.tilewidth;
```
obiot: is there not a typo though here, it should be `tileheight` ?

obiot:
I'm not sure I understand though, if `tilewidth` and `tileheight` are anyway parameters of the constructor, why add new ones in `data` ?

src/level/TMXLayer.js
YosuaHamonangan: oh yeah its a typo

YosuaHamonangan:
because the data variable contains the information for individual tile layer, so if I want to have different tile size for each layer, the best place to put the size information is in that data variable. 

and in my file format, I leave the original tilewidth and tileheight undefined

obiot:
hi, will you update the change request ? as well I think It would make more sense to prioritise `data` over the given parameters.

I'm I am correct, Tiled used to have that features where it was possible to have layers with different tile size, and if one day it comes back, then trying `data.tilexxxx` first would make it works with it too.

YosuaHamonangan:
oh, I see, make sense, will fix the request

obiot:
👍

PR #942, opened 2018-07-27
Directly defining the size of a level in tmx

YosuaHamonangan:
Since in the IDE I let the user define their own level size (they don't need to define column and row of the tile layer), I put the size in level json file in levelWidth and levelHeight

### DIFF
--- a/src/level/TMXTiledMap.js
+++ b/src/level/TMXTiledMap.js
@@ -185,7 +185,12 @@
 
             // map type (orthogonal or isometric)
             this.orientation = data.orientation;
-            if (this.orientation === "isometric") {
+
+            // manually defined level size
+            if (data.levelWidth && data.levelHeight) {
+                this.width = data.levelWidth;
+                this.height = data.levelHeight;
+            } else if (this.orientation === "isometric") {
                 this.width = (this.cols + this.rows) * (this.tilewidth / 2);
                 this.height = (this.cols + this.rows) * (this.tileheight / 2);
             } else {


obiot:
honestly, I'm not a big fan of this one :
- even though you don't use Tiled to generate the levels, if the level is tile base you still need to specify  how much tiles you have vertically (col) and horizontally (row). Also not setting these properties might cause other issue as they are being used later in the code. Plus they are easy to calculate (e.g. width / tilewidth)
- the current change most probably break potential support for non orthogonal maps (iso, hexa), as cols/rows informations are missing

YosuaHamonangan:
The reason I cannot specify the col and row is because each layer have different tilesize (with constant level size, the row and col of every layer is different). I can give col and row for individual row like previous change, but I cannot give global col and row

obiot:
No offence, but the current change request looks more like a hack on the current TMX format.

We need to find a better way to make this looks more generic and better integrated. Taking this change, and assuming you add 10 layers even with the same tile size to a level, by not specifying a tile width and height, the folllowing will always fail and return a new renderer :
https://github.com/melonjs/melonJS/blob/master/src/level/TMXTiledMap.js#L229-L241
also, this will break the isometric support, if you add it later since you then ignore the orientation.

You could simply assume that the first Tile Layer created, define the map tile height and width ? 

after all this is how it works in Tiled, and larger tiles will be aligned on the small one :
http://doc.mapeditor.org/en/stable/reference/tmx-map-format/

> The tilewidth and tileheight properties determine the general grid size of the map. The individual tiles may have different sizes. Larger tiles will extend at the top and right (anchored to the bottom left).

YosuaHamonangan:
Well I could use the first tile to define the map's tile height and width,  but I think this
https://github.com/melonjs/melonJS/blob/86e08a3fae16817770d8d09c55ef8b86f2677cd3/src/level/TMXTiledMap.js#L235-L238
would still fail in the case of different tile size and return new renderer if the current rendered layer doesn't match the size of the previous layer. So I think the best approach for multiple tile size would be to make the TMXTileMap.renderer an array of renderer. The getRenderer function will check the canRender function for every renderer created and will only create new renderer if no matching size found. 

But honestly I can't predict what other problem will be caused by this. And I haven't touch isometric map and currently have no idea about it so I don't know what will happen in isometric map. So, just want to know your opinion, do you want me to continue adding the different tile size feature and fix the problem one at a time (I'll try to make changes that doesn't effect the format from tiled) , or do you want me to use tiled format and make the IDE map editor the same as tiled?

obiot:
my recommendation is either we use the first layer as the reference tile size for the map object (which is more or less equivalent to TileId), or we just allows one tilesize per map (which is also more or less the case in Tiled, even though it will display subsequent layer with different tile size just fine).

YosuaHamonangan:
ok then, I'll adjust the IDE to be like tiled

obiot:
👍  closing this one then

PR #945, opened 2018-08-15
Shorter _super function

YosuaHamonangan:
I have a suggestion, what if we store the reference for the super class whenever an object is extended. This way, the user don't have to write the super class when calling _super function. 
So instead of :
this._super(me.Sprite, 'init', [x, y, settings]);
we can just write:
this._super('init', [x, y, settings]);

### DIFF
--- a/src/lang/jay-extend.js
+++ b/src/lang/jay-extend.js
@@ -103,8 +103,13 @@
         }
 
         // Apply syntactic sugar for accessing methods on super classes
-        Object.defineProperty(Class.prototype, "_super", {
-            "value" : _super
+        Object.defineProperties(Class.prototype, {
+            "_super": {
+                "value" : _super
+            },
+            "_superClass": {
+                "value" : this
+            }
         });
 
         // Create a hidden property on the class itself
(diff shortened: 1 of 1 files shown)


parasyte:
Have you read the original article that describes the design decisions of the inheritance stub API?

http://blog.kodewerx.org/2014/03/melonjs-should-be-all-about-speed.html

If you have found a way to keep it fast and flexible as described in that doc, then thank you! But please ensure the change does not regress any features or performance associated with typing an extra 20 characters or so.

YosuaHamonangan:
Sorry, is there anyway to measure how fast it is?

obiot:
Hi, first of all this library is actually in that repo https://github.com/parasyte/jay-extend , so any changes should rather be proposed and discussed there.

second, I'm trying to actually get rid of it by moving to ES6, so unless there is a massive improvements in terms of performances, I don't honestly want to impose this change to "our" developers, as this is somehow temporary then (at least I hope) and to be replaced by a more standard [approach](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes)

furthermore, I'm trying to adopt a more progressive approach in transitioning to ES6, so my next step I think is to actually use the ES6 class definition, but use Babel to transpile the whole thing back to ES5.

anyway, I will close this PR. Feel free to re-open it on the link here above.

parasyte:
@YosuaHamonangan I initially just used the particle emitter for testing object creation speed. You can see some examples and stuff here: http://blog.kodewerx.org/2014/10/melonjs-should-be-all-about-speed-part-3.html

PR #946, opened 2018-09-04
Remove adjust position for non-tiled map

YosuaHamonangan:
The object position for isometric map in melon editor and in melonJS is already the same, unlike in tiled, so there is no need to adjust position. 
For now, I use data.tiledversion in the map object to check if a map is from tiled

### DIFF
--- a/src/level/TMXObject.js
+++ b/src/level/TMXObject.js
@@ -186,7 +186,9 @@
             }
 
             // Adjust the Position to match Tiled
-            map.getRenderer().adjustPosition(this);
+            if (map.isTiled) {
+                map.getRenderer().adjustPosition(this);
+            }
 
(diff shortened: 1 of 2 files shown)


obiot:
Hi Yosua,

Unfortunately the `tiledversion` property is only available since Tiled version 1.0.1, which means that this changes would break any map exported with a previous version.

see here :
http://doc.mapeditor.org/en/stable/reference/tmx-map-format/

Why not doing thing the opposite way and add a specific property for map built manually ? (like "custom:true" or something similar)

obiot:
Or just simple add an “editor” property that would contains “melonjs” in our case, then you can check for it.

obiot:
yeah, now I like it ! :P

PR #957, opened 2018-11-05
Fix bug for flip image in trimmed texture

YosuaHamonangan:
There's a bug when a texture packed trimmed image is flipped, the anchorpoint of the image stays the same which causes the image to move


obiot:
Hi, sorry, but I'm don't really agree with the latest changes, why do we have to add another Vector object "flippedAnchorPoint" that is specific to the flip method ? we already have property that indicated is the image was trimmed, the size, the anchorPoint, this should be enough to to make this thing works without adding yet another object to the atlas.

YosuaHamonangan:
I added the latest one because the previous one doesn't work on a regular sprite sheet. For regular sprite sheet the anchor point shouldn't be changed, but I'm not sure how to check if the frame object come from a spritesheet or trimmed texture atlas. can you give suggestion better way to differentiate this two cases?

obiot:
texture atlas has a `trimmed` boolean property, that one could be added to the frame definition passed to melonJS and then you can checked if `trimmed` true or not and change the anchorPoint accordingly, that sounds to me more like a generic fix for this.

YosuaHamonangan:
ok, how about like this

obiot:
yes, much nicer, and you see much more simple :)

PR #958, opened 2018-11-08
Fixed for Wechat mini game platform.

qichunren:
There is no "document.location" for Wechat mini game platform.

image


obiot:
Thanks !

PR #962, opened 2018-12-07
fix container clipping bug

YosuaHamonangan:
the currect clipping doesn't clip the correct region because it clips at two times the actual position, for example when x=100 the clipping start at x=200. I think the clipping should be done 1st before translating the renderer

### DIFF
--- a/src/renderable/container.js
+++ b/src/renderable/container.js
@@ -854,9 +854,6 @@
 
             this.drawCount = 0;
 
-            // adjust position if required (e.g. canvas/window centering)
-            renderer.translate(this.pos.x, this.pos.y);
-
             // clip the containter children to the container bounds
             if (this._root === false && this.clipping === true && this.childBounds.isFinite() === true) {
                 renderer.clipRect(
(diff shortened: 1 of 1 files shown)


obiot:
can you detail your use case, or share an example showcasing your issue ? because if I take the Shape example (and enable clipping for both child container), applying your change will break the clipping.

YosuaHamonangan:
I found the problem when I add a sprite to a clipped container like this

       var spriteSettings = {
            image: 'player75x96',
            framewidth: 75,
            frameheight: 96,
            anchorPoint: {x: 0, y: 0}
        }
        var posX = 50;
        
        var container = new me.Container(posX, 0, 100, 100);
        container.anchorPoint.set(0, 0);
        container.clipping = true;
        container.addChild(new me.ColorLayer("", "blue"));
        container.addChild(new me.Sprite(0, 0, spriteSettings));
        me.game.world.addChild(container);
        
        me.game.world.addChild(new me.Sprite(posX, 100, spriteSettings));

when posX is 0, the clipping is correct, but when the posX is increased, the left side of the sprite is clipped by the same amount of posX

PR #964, opened 2019-01-24
add name to sprite settings

setsujinteki:
Hello Mr. Oliver

I'm Christian from melongamings. I add name to sprite setting because i need it for the editor.

### DIFF
--- a/src/renderable/sprite.js
+++ b/src/renderable/sprite.js
@@ -35,6 +35,8 @@
     {
         /** @ignore */
         init : function (x, y, settings) {
+            // layer name
+            this.name = settings.name || "me.Sprite";
 
             /**
              * pause and resume animation


obiot:
hi, ok yes, we can definitely define `name` at a higher level since mostly all objects inheriting from sprite class are re-defining a `name` property.

however, I will add it later by myself, as I want to add this to me.Renderable directly and not me.Sprite, and anyway this change is not properly implemented (there is no documentation for the new name property, and the comment line above is mentioning me.Layer instead of me.Sprite).

last but not least, `name` should be for information only, if you need to detect an object type you should rather use `instanceof` (e.g. `if (myObject instanceof me.Sprit) then....`)

obiot:
@setsujinteki here it is : https://github.com/melonjs/melonJS/commit/7606148345c2d83022a63a03fcabf063ed711c86

PR #965, opened 2019-02-01
Simple 1 line recommendation for initial documentation

dwkrueger2:
I think its a good idea to point people to the over view first.
I think you should also separate 'Building MelonJS' instruction from 'Using MelonJS' 
That would really help clarify things.


dwkrueger2:
I added a tutorial to the examples directory.  It's a bit Simpler than all the others and I think provides a better basic starting point.  As my Project Continues I'd like to add to the tutorial as a series.

obiot:
Hi, thank you very much ! I like the idea of having a more simple tutorial, not sure the examples folder here is THE right place to put it though. I'm going to merge it for now, we can see later about moving to it's own repository (like for the 2 others tutos)

PR #968, opened 2019-02-15
Updated comments of setInterval and setTimeout

dwkrueger2:
Functionality of the functions was unclear.  Updated documentation to more accurately describe their function


dwkrueger2:
All commits are improvement in documentation.  I really think the accel property should be marked as experimental since it does not properly calculate velocity increase.

PR #970, opened 2019-02-18
Updated Documentation

dwkrueger2:
I was expected that if you added this object to a container that the container would resize the render.  It does not.  It fills the whole viewport with the color


obiot:
yeah correct, although this should be modified to only fill the holding container, or at least the size of the current viewport.

PR #975, opened 2019-03-09
Corrected incorrect pointer offset - Please check - Critical

dwkrueger2:
This is a potentially critical problem.  The click was off on an UI click events.. the previous setting this.pos.x  registered to the incorrect location.  I could not exactly nail down which coordinate this.pos.x.  I think it was the clientX which seemed to be consistently different from the canvas element position which is given by either event.layerX/Y  or event.offsetX/Y... So it seems that the clientXY is different from the canvas application position.  I tested this in my application and it seems to work well.

### DIFF
--- a/src/input/pointer.js
+++ b/src/input/pointer.js
@@ -245,8 +245,8 @@
 
             this.type = event.type;
 
-            this.gameScreenX = this.pos.x;
-            this.gameScreenY = this.pos.y;
+            this.gameScreenX = event.layerX;
+            this.gameScreenY = event.layerY;
 
             // get the current screen to world offset
             if (typeof me.game.viewport !== "undefined") {


dwkrueger2:
I'm rethinking the problem is not where I corrected it.  Attache is a screen shot of the problem.  
The Top left corner of the 'P' in Play should be at 172,130 but the cursor which is reported is in the lower left at 170,99

When corrected it earlier I had the auto scale turned off. But this screen shot is from the Canvas settings as follows:
 if (!me.video.init(AppJS.contentWidth, AppJS.contentHeight, {wrapper: "screen"
         , scale:'auto', scaleMethod: "stretch", doubleBuffering :true}))

The scaling seems to be pushing the Renderable up and not position where I am calling for it to be positioned (which is the 172,130).
This makes me think that the position of the renderable is not getting updated when the image is scaled.

I can move this to the form if you want.


image

dwkrueger2:
Okay One more comment... I noticed that my Vertical cursor reporting was limited to 800 and suspected a scaling issue.  But I had set the screen container and the canvas element CSS size programatically with 
  $('#screen').height(AppJS.contentHeight);
   $('#canvas').height(AppJS.contentHeight);

However, the CSS 
#screen canvas { /* coppied from the exampes/platformer */
    width:1024px;
    height: 1024px;  /* <-- when it was 800 I got the scaling issue */
    margin: auto;
--
This threw me for a loop because I thought the CSS would have been applied prior to getting to this point in the code.  This might be a way that the Chrome, Inspector is applying the CSS.  Because if you make changes while coding to the CSS chrome will automatically adjust the CSS... Not caring if the CSS was changed programatically.

Regardless:  When the CSS is set to match the Canvas Element created in the me.video.init there is still a clickable offset that does not match.  The original correction proposed does fix this error.

obiot:
would you be able to share your game so that I can give it a try ?

else in terms of proposed changes, firs thing is that is should rather be changed on those lines :
https://github.com/melonjs/melonJS/blob/master/src/input/pointer.js#L212-L218

and especially here : https://github.com/melonjs/melonJS/blob/master/src/input/pointer.js#L218
as this is where the mouse/pointer input is taken and then transformed internally .

now I would not use layerX/Y as this is not a standard property (see the MDN page), but you could try to rather pass pageX/Y to the me.input.globalToLocal function (on line 218) and see if it fixes things,

the other point might be related to bounding box, anchorpoint and even detection where they don't always act as best buddies when a renderable use an Anchor Point different than [0, 0], so something else to try as well.

obiot:
@see https://github.com/melonjs/melonJS/issues/976

Issue #925, opened 2018-05-16
Tween API improvements


obiot:
The Tween API could use some improvements, allowing to pass more standard parameters through the constructor rather than just the duration (like repeat, easing function, yoyo, etc...)

now :
```javascript
tween = new me.Tween(myObject.pos).to(
    {y: 200}, // properties
    3000 // duration
).onComplete(myFunc);
tween.easing(me.Tween.Easing.Bounce.Out);
tween.repeat(10);
tween.yoyo(10);
tween.start();
````

proposition :
````javascript
tween = new me.Tween(myObject.pos).to(
    {y: 200}, // properties
    { 
        duration : 3000,
        easing: me.Tween.Easing.Bounce.Out
        repeat : 10,
        yoyo : true
    } 
).onComplete(myFunc);
tween.start();
````

obiot:
thanks to @0xf0f0f0 for this one !

PR #1020, opened 2021-07-06
Tween API improvements #925

0xf0f0f0:


obiot:
Hi Thank you very much for the Pull Request, but the changes are breaking backward compatibility, and makes it not anymore aligned with the usual Tween implementation out-there.

I understand the changes, but why not check for the type fo the second parameter, and if not an integer (and object instead) than assume this is directly the list of properties ?

obiot:
@0xf0f0f0 Hi, sorry if took us so long, but there were some pending large changes (see the last commits), but now that they are done, I merged your pull request, and improved it as well to make it backward compatible,

thanks again !

PR #1059, opened 2021-11-05
Change enablePointerEvent() signature and promote to public function

kkeiper1103:
Currently, `me.input.bindPointer(me.input.pointer.LEFT)` relies on `me.input.pointer` being initialized via `enablePointerEvent()`. This leads to calling functions specifically for their side effect of calling `enablePointerEvent` via `registerPointerEvent`

Example of previous behavior:

## title.js
```
onResetEvent() {

    // this calls enablePointerEvent(), which is what we *really* want. We don't care about the pointerdown event itself
    me.input.registerPointerEvent("pointerdown", null, null);

    me.input.bindKey(me.input.KEY.ENTER, "enter", true);
    me.input.bindPointer(me.input.pointer.LEFT, me.input.KEY.ENTER);

    this.handler = me.event.on(me.event.KEYDOWN, function (action, keyCode, edge) {
	if (action === "enter") {
		me.state.change(me.state.PLAY);
	}
    });
}
```

Example of new behavior / usage:
## game.js
```
onload() {
    // Initialize the video.
    if (!me.video.init(640, 480, {parent : "screen", scale : "auto", scaleMethod : "flex-width"})) {
        alert("Your browser does not support HTML5 canvas.");
        return;
    }

    // initialize pointer events
    // please notice that the signature changed! This is because it registers more than just one type of event
    me.input.enablePointerEvents();

    // ...
},
```

## title.js
```
onResetEvent() {
    me.input.bindKey(me.input.KEY.ENTER, "enter", true);
    me.input.bindPointer(me.input.pointer.LEFT, me.input.KEY.ENTER);

    this.handler = me.event.on(me.event.KEYDOWN, function (action, keyCode, edge) {
	if (action === "enter") {
		me.state.change(me.state.PLAY);
	}
    });
}
```

### DIFF
--- a/src/input/pointerevent.js
+++ b/src/input/pointerevent.js
@@ -14,7 +14,6 @@ import Container from "./../renderable/container.js";
 import Renderable from "./../renderable/renderable.js";
 import { world, viewport } from "./../game.js";
 
-
 /**
  * A pool of `Pointer` objects to cache pointer/touch event coordinates.
  * @type {Array.<Vector>}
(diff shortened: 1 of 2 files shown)


obiot:
fix is in ! 
https://github.com/melonjs/melonJS/commit/3dae2367155bbdf15e5481dd3a634429563155ae

It will be part of the 10.1.0 release, that should be out in a couple of days max, I'm working on some improvements on the WebGL renderer (to further optimise it) and I'm almost done. in the mean time, just initialise first an event in the tutorial to avoid the bug/regression :)

thanks again for reporting this one, I added your name in the Changelog :)

PR #1080, opened 2022-03-02
Changed this.releasePointerEvent to releasePointerEvent as it was causing a "this is undefined" error.

siauderman:
Dear Olivier, I believe this fixes the bug. I have checked releasePointerEvent and it does not refer to "this" at all. Cheers, Wilson.


obiot:
good catch, we still have a few "unfound" bugs here and there following the code rewriting to ES6 :) ! thank you so much !

PR #1099, opened 2022-06-07
improve the build scripts to prevent accidental publishes without testing

L1lith:
By overriding publish command (using "prepublish") instead of using the release script we can prevent accidentally doing the publish command without testing (as it's now enforced) and the command is more conventional


obiot:
indeed I did have a couple of "failed" publish in the past ! that's a good suggestion, thanks for that

Issue #1089, opened 2022-03-19
Replace JSDoc as it is not maintained and is legacy software


ShukantPal:
Hi folks, JSDoc is [no longer being actively maintained](https://github.com/jsdoc/jsdoc/commits/main) and that's why I built a better, maintainable toolset called [webdoc](https://www.webdoclabs.com?utm_campaign=melonjs&utm_source=ni&utm_medium=GitHub). webdoc has a more [modern theme](https://webdoc-labs.github.io/example-documentation/index.html) that's fully customizable! I would be down to helping you move your documentation to webdoc, the choice is up to you.

I think webdoc will bring a lot of clarity to your existing documentation, and it is pretty good for this use case. See the PixiJS (also a renderer) documentation site: https://api.pixijs.io

obiot:
hi, thanks for reaching out ! 

we have been discussing this actually for years (literally #497) but nobody never took on the task, as honestly it does look like a big one, maybe not anymore in "converting" the tags, but especially in recreating a specific template/theme and all.... but definitely, If you are willing to help, I could hardly say now :)

ShukantPal:
Ok will take a look next weekend.

obiot:
Hi @ShukantPal, I went ahead and add some first npm task to generate webdoc documentation, it's as straightforward as you were implying though.

So first comments and related questions :
- I literally have a billions of `[...] does not have a "-" token preceeding description [...]` errors  : is there a way to configure this ? this was not an issue with jsdoc, and I really don't feel like going through all the source code to add a `-` between each function parameter and their description

- is it possible to output the line number where the tag parsing error is coming from ?

- I use this "trick" (https://github.com/melonjs/melonJS/blob/master/src/geometries/line.js#L15-L31) to generate different doc entry for the same function when the function takes different parameters, but webdoc does not seem to like it (e.g. `"x" is not a parameter & cannot come after the last parameter "`), is there a way to disable warning or force it, similarly to eslint for example ?

- what are the `[DepsChain] xxxx is a member of xxxx` output for ?

- every time I run webdoc it creates an `example.api.json` file at root level ?

thanks !

Note: I'm using the default template for now ("@webdoc/default-template"), config file is here : https://github.com/melonjs/melonJS/blob/master/webdoc.conf.json

ShukantPal:
* I'll create an option to disable these warnings & try to release this weekend. I don't think there's a way to disable it immediately.
* Can you post which error you're seeing in this case?
[...]

PR #1100, opened 2022-06-07
Prune out needed @function tags & convert required tags to @method

ShukantPal:
- Prune out unneeded @function tags and convert rest to @method
- Bump to webdoc 1.6.1

See #1089


obiot:
let me bump the version to 10.11 first :)

PR #1101, opened 2022-06-07
Fix all DocParser errors in webdoc

ShukantPal:
* Use @member instead of @type so webdoc knows we're documenting a class member.
* Change `/**` to `/*` where it's not an actual _documentation_ comment.
* Fix the "device" namespace so the namespace is the actual object and remove all @memberof tags pointing to that 
namespace in the object properties (webdoc parses the object properties automatically!)≈


ShukantPal:
See #1089

PR #1102, opened 2022-06-07
fix a bug with the new script making install fail

L1lith:
I made a mistake. The "prepublish" script runs before publish and install, but "prepublishOnly" only works on publish


obiot:
@L1lith I'm not sure which one is better, now if I run a `npm install` to update outdated dependencies, after actually updating the required packages, it just execute everything ; build melonJS, generate the dist files, and run the test suite. is this how its supposed to be ? I mean I don't somehow mind, it's not like external dependencies are being updated all the time, but it seems a lot.

Issue #276, opened 2013-09-09
Text Input


Zaraka:
I need to somehow implement console + chat into my game. But I'm struggling with TextInput. MelonJS doesn't provide any solution for this problem. I have tried emulate it through Keyboard events, but it's terrible (You can check it at my Mana repo). Is there any better solution?

PR #1119, opened 2022-08-11
New UI components

wpernath:
Hi,

as discussed in another thread, here are two components you can use as you want:
- BaseClickableContainer: The base implementation of something clickable (inspired & copied from GUI_Object)
- BaseTextButton: A round rectangle based, hover supported, text button implementation as one implementation of BaseClickableContainer

Thanks a lot for all your help!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,15 @@
 
 ## [13.2.0] (melonJS 2) - _2022-xx-xx_
 
+### Added
+- Geometry: new base Point geometry object
+- Tiled: now properly support Point geometry if defined in a map
+
 ### Changed
-- Core : replaced internal reference to the deprecated `trim[Left/Right]` methods by their `trim[Start/End]` equivalents
+- Core: replaced internal reference to the deprecated `trim[Left/Right]` methods by their `trim[Start/End]` equivalents
+
+### Fixed
+- WebGL: fix a regression with global opacity not being properly cascaded to texture in WebGL mode (thanks @wpernath)
 
 ## [13.1.1] (melonJS 2) - _2022-08-10_
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "melonjs",
-  "version": "13.1.0",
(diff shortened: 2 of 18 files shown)


obiot:
thinking about renaming everything as well, what about :
- UIBaseContainer 
- UISpriteContainer (former GUI_Object)
- UITextButton

what do you think ?

wpernath:
Good idea! Let’s do it. 

Just forward thinking: what about a
- UIListView (a scrollable Container with any content?)
- UIImageTextButton
- UIcheckBox

As I am new in JavaScript etc., but I personally don’t like to use a html popup for letting the user input some text. But I don’t know how complicated it is to provide:

- UITextInput (for example to let the user input his name etc.)


Best,
  Wanja

wpernath:
I have created a new folder in `src/renderable` named `ui` and renamed the two classes accordingly.

obiot:
@wpernath later than never, this is finally merged and all ! thanks again for the PR.

How is your game coming along ?

wpernath:
Good idea! Let’s do it. 

Just forward thinking: what about a
- UIListView (a scrollable Container with any content?)
- UIImageTextButton
- UIcheckBox

As I am new in JavaScript etc., but I personally don’t like to use a html popup for letting the user input some text. But I don’t know how complicated it is to provide:

- UITextInput (for example to let the user input his name etc.)


Best,
  Wanja

obiot:
the `UITextInput` is definitely not a good idea in my opinion, as implementing a text input component that support different languages and all is a full project by itself. And then there is the whole keyboard rendering itself, that anyway is project specific, so nobody never agreed that having a base "generic" keyboard was marking sense.

we discuss this multiple time over the course of the last few years, and the conclusion was always that using css was just easier, I could not find back all the post on the topic, but here are a few of them  :
https://github.com/melonjs/melonJS/wiki/Frequently-Asked-Questions#form_inputs
https://github.com/melonjs/melonJS/issues/276
https://www.html5gamedevs.com/topic/34357-do-you-build-your-uis-in-melonjs-or-dom/#comment-197401

PR #1128, opened 2022-08-26
Put the new tutorial in the README

L1lith:


obiot:
indeed, but I will also remove the old one anyway

PR #1140, opened 2022-10-03
Update the tests to be browser based and use Mocha

L1lith:


obiot:
what? wow you did all the migration to mocha !

PR #1174, opened 2023-01-26
Improve documentation for addTransition().

TOertel:
When trying to understand transitions, the documentation for addTransition() was ambiguous as to whether the transition applied when leaving FROM the given state, or going TO the given state.  Updated documentation to clarify (after looking at code).


obiot:
thank you ! Much appreciated :)

PR #1183, opened 2023-04-19
fix typo of document for checkVersion()

orcation:
Fix typo in parameter descriptions.


obiot:
good catch, thank you !!!

PR #1200, opened 2023-08-17
fix: CanvasRender options typo

0o001:


obiot:
wow good catch, I cannot believe it was working with this type !

PR #1204, opened 2023-10-10
feat: handle assets load error

z0mb1:
- [x] add event for displaying a error during loading `LOADER_ERROR`
- [x] add function `rePreload` to load asset, that was loaded with error

Userstory: I am trying to load project, but due to problems with the internet connection some of my assets load fails. I don't want to reload page and load whole project again. I want to upload again only failed assets.

Example decision: add notification about load failure with button to upload this asset again using `rePreload` function

### DIFF
--- a/src/loader/loader.js
+++ b/src/loader/loader.js
@@ -46,6 +46,19 @@ export let onload;
 export let onProgress;
 
 /**
+ * onError callback<br>
+ * each time a resource loading is failed, the loader will fire the specified function,
+ * giving the actual progress [0 ... 1], as argument, and an object describing the resource loaded
+ * @default undefined
+ * @memberof loader
+ * @type {function}
+ * @example
+ * // set a callback for progress notification
+ * me.loader.onProgress = this.updateProgress.bind(this);
+ */
+export let onError;
+
+/**
  * list of parser function for supported format type
  */
 let parsers = new Map();
@@ -60,6 +73,10 @@ let resourceCount = 0;
 let loadCount = 0;
 let timerId = 0;
 
+/**
+ * Assets uploaded with an error
+ */
+const failureLoadedAssets = {}
 
 /**
  * init all supported parsers
@@ -111,6 +128,7 @@ function checkLoadStatus(onloadcb) {
  * @ignore
  */
 function onResourceLoaded(res) {
+    delete failureLoadedAssets[res.src]
     // increment the loading counter
     loadCount++;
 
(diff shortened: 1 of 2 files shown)


obiot:
Hi, I like this idea of indeed being able to retry preloading assets in case the network is not stable.

Just one thing not clear in the example you provided in the source code where to you put the below code ? My only concern is about this feature being a user choice. what I mean is that by default we still have the current behaviour, but then you can choose to try re-preloading certain element.

````javascript
@example 
   event.on(
       event.LOADER_ERROR,
       (res) => {
           // custom function
           showErrorNotification({
               text: `Error during loading content: ${res.name}`,
               done: loader.rePreload(res.src);
           })
       }
   );
````

z0mb1:
> Hi, I like this idea of indeed being able to retry preloading assets in case the network is not stable.
> 
> Just one thing not clear in the example you provided in the source code where to you put the below code ? My only concern is about this feature being a user choice. what I mean is that by default we still have the current behaviour, but then you can choose to try re-preloading certain element.
> 
> ```js
> @example 
>    event.on(
>        event.LOADER_ERROR,
>        (res) => {
>            // custom function
>            showErrorNotification({
>                text: `Error during loading content: ${res.name}`,
>                done: loader.rePreload(res.src);
>            })
>        }
>    );
> ```

Hi. We could put the code itself in the `onload` function, or any other function that is called when the application is initialized. my example is really quite specific. But this is due to the fact that the processing of unloaded assets can be implemented at the discretion of the user. In my example, as in my project, the function calls a notification with a button to reload. But you can do it in another way. For example, on a custom loading screen, display all unloaded assets. Also, I didn't use the standard UI, and I can assume that the implementation of this function will depend on the frameworks and libraries used. I would appreciate it if we could come up with a more understandable example together.

Снимок экрана 2023-10-11 в 14 42 27

obiot:
Hi Sorry, for the late feedback, and got it, and I will definitely give it a try once merged and think about how I would use it (to come up with an example)

Last comment or request before merging, can we maybe rename `rePreload()` to `reload()` ? I just feel the name `rePreload` is kind of strange :P

z0mb1:
> Hi Sorry, for the late feedback, and got it, and I will definitely give it a try once merged and think about how I would use it (to come up with an example)
> 
> Last comment or request before merging, can we maybe rename `rePreload()` to `reload()` ? I just feel the name `rePreload` is kind of strange :P

sure, done

obiot:
merged ! :)

thank you so much again for your contribution !

PR #1205, opened 2023-10-10
prevent load previously loaded resource

z0mb1:
Userstory: I have several levels with separate manifests. This manifests contains some part of the same assets. If I load second level after complete first, I don't want to load asset I previously loaded.

- [x] check if asset has already loaded and skip download


obiot:
love it ! thank you :)

obiot:
Hi @z0mb1 one more thing, before submitting a PR, please do make sure to build the whole library to avoid linting issues or even regression.

obiot:
version 15.14 published !

PR #1206, opened 2023-10-16
add types for error handle

z0mb1:
I forget to add types to my previous PR


obiot:
you actually don't need, types are automatically generated when building the distribution files as part of the release. I will publish the new version today, I was trying to squeeze in something else, but I cannot find the time to finish it these days so I 'll keep it for a next version :)

Issue #1199, opened 2023-08-08
this.body is pointing to Dom type when using Typescript Boilerplate


brunofunnie:
I notice that when using the Typescript boilerplate and when setting a new Body it refers to the DOM instead of MelonJS.

The code:
![image](https://github.com/melonjs/melonJS/assets/5279392/f136ecbe-eb69-4a7b-9a2d-be05d7b897d8)

On VSCODE shows:
![image](https://github.com/melonjs/melonJS/assets/5279392/0ac57261-d4d8-4bc6-975c-e6b17a49752f)

Typescript reference of Body:
![image](https://github.com/melonjs/melonJS/assets/5279392/e1a9d3b7-d0e1-4010-9647-5104cf2bd81e)

Using a just cloned Typescript Boilerplate and following the https://codemonkey.club/getting-started-with-melonjs/

obiot:
thank you very much for the report, we will look into it. 

We are still in the process of providing proper typings for melonJS, please do share if you find any other issues. In the mean time, you can use `@ts-ignore` to remove the warning/error report in VSC.

muyaszed:
I have a fix for this, just need the suitable replacement name for `Body` class. Atm I'm using `BodyObject` for it. the Method will still be using `body`

muyaszed:
Do I need to commit the updated Doc as well?

daniel-centore:
For now I'm working around this with:

```typescript
import { Body } from "melonjs";

const body = this.body as unknown as Body;
//  use `body` wherever needed
```

Would be great to get a real fix for it!

obiot:
me too It has been bothering me, but I really don't know how to fix it properly (I mean without renaming the object!)

obiot:
This one should be fix now, I just simply "tricked" it by importing Body within the Renderable class (together with referencing the Body class, so that the bundler doesnt complain).

I found out another issue is that rollup does not always respect the right module order in the bundle if I add import/reference for Application. Another issue, but at least the Body one should be resolved now.

I will leave this one open until the next 17.1 version is released and the fix confirmed to work.

obiot:
17.1.0 released, if you guys could confirm it's now correct from your side as well, thanks !

obiot:
closing this one, working on my side.

PR #1208, opened 2023-11-02
Change Body Class name

muyaszed:
##### Description of change
Change the Body class name to BodyObject

Please refer issue -> https://github.com/melonjs/melonJS/issues/1199

##### Merge Checklist

- [x] Build process passed (`npm run build`)
- [x] Lint process passed (`npm run lint`)
- [x] Tests passed (`npm run test`)

### DIFF
--- a/dist/types/physics/body.d.ts
+++ b/dist/types/physics/body.d.ts
@@ -3,7 +3,7 @@
  * a Generic Physic Body Object with some physic properties and behavior functionality, to add as a member of a Renderable.
  * @see Renderable.body
  */
-export default class Body {
+export default class BodyObject {
     /**
      * @param {Renderable|Container|Entity|Sprite|NineSliceSprite} ancestor - the parent object this body is attached to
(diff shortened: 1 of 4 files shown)


obiot:
hi @muyaszed appreciating the pull request, but we cannot merge this, as It will break backward compatiliby and this "just" to fix a typing issue with TS.

is a proper way not to add proper typing for [Body](https://github.com/melonjs/melonJS/blob/master/src/physics/body.js) in a dedicated .ts file, that would then be added as a source directory when generating the .ts file(s) ?

something like :

````javascript
declare class Body {
    constructor(ancestor: Renderable, shapes: Polygon, onBodyUpdate?: Function);

   ancestor: Renderable;
   bounds: Bounds;
   ...

   setStatic(isStatic?: boolean): this;
   addShape(shape:Polygon) : number;
  ...

}
````

muyaszed:
Yeah sure no problem this is just to start something which we could progress towards the better solution. I would take a look at the suggested code. Thanks for the reply

PR #1222, opened 2024-03-26
Fix Camera2d.WorldToLocal to translate game.world.pos also

pathak-prashant:
##### Description of change
WorldToLocal applies transformation on the passed vector if the camera has transformation applied to it, however it does not apply it to the game.world.pos which is added to get the file result. This results an incorrect world to local translation. This change copies the world pos vector to a local variable and applies transformation to it, before adding it to the vector.

##### Merge Checklist

- [ ] Build process passed (`npm run build`)
- [ ] Lint process passed (`npm run lint`)
- [ ] Tests passed (`npm run test`)

### DIFF
--- a/src/camera/camera2d.js
+++ b/src/camera/camera2d.js
@@ -636,8 +636,8 @@ export default class Camera2d extends Renderable {
         v = v || pool.pull("Vector2d");
         v.set(x, y);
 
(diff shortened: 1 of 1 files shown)


pathak-prashant:
Cannot complete merge checklist as unable to run npm run lint because of attached error,

image

PR #1228, opened 2024-05-15
Use tmxData.type instead of file extension and also return a result for xml extension

customautosys:
Dear Olivier,

I would like to suggest this change as for those of us using a build step (React / Vue), "tsx" is usually recognised as a typescript-jsx file type which renders us unable to import an asset with that extension as it would be parsed wrongly. Therefore, I suggest to use the type attribute instead of the file extension if available. Also, if the type is XML, it should allow the user to deal with whatever was parsed rather than simply returning nothing.

### DIFF
--- a/src/loader/parsers/tmx.js
+++ b/src/loader/parsers/tmx.js
@@ -66,18 +66,8 @@ export function preloadTMX(tmxData, onload, onerror, settings) {
                     const xmlDoc = parser.parseFromString(response, "text/xml");
                     const data = TMXUtils.parse(xmlDoc);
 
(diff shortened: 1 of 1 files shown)


obiot:
it could even be simplified to `result = data.map || data.tilesets[0];`.

after all tsx files only defines `tileset`, so if `map` is defined its a tmx file, and if not it's automatically a tsx file.

customautosys:
I just pushed a new commit but it's not appearing yet, I did ```result = data.map || data.tilesets[0] || data;``` so in case for some reason there is no "map" or "tilesets" it'll still return something just in case.

PR #1230, opened 2024-06-22
Fix running of tests on windows machine

hornta:
- Fixed so that Windows users can run the test suite.  
- Added a GH workflow that runs the tests in both node 20(lts) and node 22 and in both ubuntu and windows environment.
- Pinned version of package manager to the npm version of the npm version provided in the latest node lts version to ensure better reliability during installation of dependencies.

##### Merge Checklist

- [x] Build process passed (`npm run build`)
- [x] Lint process passed (`npm run lint`)
- [x] Tests passed (`npm run test`)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,19 @@
+name: Test
+on: [push]
+jobs:
+  test:
+    strategy:
+      matrix:
+        node: [20, 22]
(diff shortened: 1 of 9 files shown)


obiot:
that's an awesome contribution, thanks for it, will look at it later once back and merge 👍

Issue #1225, opened 2024-04-18
[documentation] migrate to vitepress-jsdoc


obiot:
with [webdoc](https://www.webdoclabs.com) not being supported anymore, proposal is to migrate to [vitepress-jsdoc](https://blakmatrix.github.io/vitepress-jsdoc/)

obiot:
or typedoc now that (since version 17.4.0) we have proper TS definitions/declarations

PR #1232, opened 2024-06-23
Replace webdoc with typedoc

hornta:
closes https://github.com/melonjs/melonJS/issues/1225

- Replaced webdoc with typedoc
- Removed  and git-ignored `docs/` from the repository.
- Added a GitHub workflow to build an artifact and attempt to deploy that artifact to GitHub Pages.
- Better enforced line break style. Now the linter scripts checks for line break style and is aligned with the .editorconfig file. I replaced *some* deprecated stylistic rules with the new stylistic eslint plugin. https://eslint.org/blog/2023/10/deprecating-formatting-rules/

I've tested that links from the docs into the source code. I've also made sure the documentation images appears in the docs. typedoc only seem to support relative paths to images in jsdoc so I had to inline the images under the src folder, however I put them in an images folder so that they are clearly separated from the actual source code. 
I've bundled changes affecting many files into commits to ease the review process.


eslint.config.mjs
```diff
@@ -16,29 +18,25 @@ export default [
         name: "eslint/global-rules",
         languageOptions: {
             "ecmaVersion" : 2020,
```
obiot: @hornta  any reason why you replaced the previous emacVersion 2022 by 2020 ?
I did changed back to 2022 while resolving conflicts, but if there is an actual solid reason, happy to put back 2020

hornta: Not sure why I changed it. Good you caught that 👍

PR #1233, opened 2024-06-23
Source code formatter

hornta:
This PR introduces code formatting. It's like a linter but focuses more on how the code looks, rather than how it functions. The purpose of this change is to have a set way of formatting the source code so that new contributors can focus on features and bug fixes, and less time on things like questioning if some code is "correctly" formatted or not. Currently there is the chance of new contributors getting discouraged from making changes because there is no formatter. 

>  Personally, I rely on a formatter being in place. It lets me just write whatever code I want, ugly or not, but I know when I save the files, it prints it nicely for me and saves me tons of keystrokes otherwise put on aligning, indenting and what not.

This PR relies on https://github.com/melonjs/melonJS/pull/1230 and my plan is to run the formatting check of the code in the GitHub workflow, just as it does with the linting.

I've added some default workspace vscode settings so new contributors using vscode get automatic formatting in javascript files upon save. For contributors NOT using vscode, I've added a precommit hook that is installed upon running npm install using https://github.com/evilmartians/lefthook, which is formatting the files correctly just before commiting.

The formatter tool is provided by https://biomejs.dev/. It also comes with a linter(a replacement for eslint), and a tool that organizes imports according to some rules. I've disabled those to minimize changes and also because eslint is already in place here.

Another small upside with this change is that indentation is changed from spaces to tabs, letting us customize and view the code in whatever tab size is configured. I don't think many people know of this, but GitHub lets you customize how many spaces a tab should be visualized in. I personally have it configured as 2 spaces. 
image


obiot:
@hornta few conflicts to be resolved before I merge, and the "resolved conflicts" button is greyed out on my side :)

obiot:
also, I bump the current (dev) version to 17.5, and updated the CHANGELOG, you might have to fetch those changes as well on your side

hornta:
> @hornta few conflicts to be resolved before I merge, and the "resolved conflicts" button is greyed out on my side :) Screenshot 2024-06-24 at 10 40 34 AM

I've resolved the conflicts now.

PR #1248, opened 2025-06-27
Melon editor dev/web gl rendering fix

Vareniel:
##### Description of change
- Updated the way the Quad Compositor renders when updating textures repeatedly.
- Fixed color blending issue in CanvasRenderer when using multiple Light2D objects at the same position. Previously, when multiple Light2D objects were placed at the same position using the CanvasRenderer, the colors didn't blend correctly. This issue has now been resolved, and Light2D objects properly blend colors in the CanvasRenderer. **Explanation**: In the CanvasRenderer context, the first Light2D object will set the blend mode to this.blendMode. However, subsequent Light2D objects do not reset the blend mode after calling ctx.save() and ctx.restore(). This was causing the incorrect color blending behavior. If ctx.save() and ctx.restore() were not used, the blend mode would only need to be set once, and the blending would work as expected.

##### Merge Checklist

- [ x ] Build process passed (`npm run build`)
- [ x ] Lint process passed (`npm run lint`)
- [ x ] Tests passed (`npm run test`)

### DIFF
--- a/packages/melonjs/src/renderable/light2d.js
+++ b/packages/melonjs/src/renderable/light2d.js
@@ -175,6 +175,17 @@ export default class Light2d extends Renderable {
 	}
 
 	/**
+	 * preDraw this Light2d (automatically called by melonJS)
+	 * Note: The renderer should set the blend mode again (after drawing other Light2d objects)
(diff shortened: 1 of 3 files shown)


Noodles006:
收到您的来信，谢谢！

PR #1533, opened 2026-07-01
feat(shader): ShaderEffect.setTime + setTexture (extra samplers, closes #1532)

### LINKED ISSUE #1532: [Feature request] Allow engine-managed extra textures for custom shaders
**Is your feature request related to a problem? Please describe.**

I’d like to use NoiseTexture2d or other engine textures as extra samplers in custom shaders. Right now it works nicely as Sprite.image or normalMap, but for custom shader effects like distortion I still need to manually create/bind WebGL textures.
**Describe the solution you'd like**

A small engine-managed API for passing extra textures into custom shaders, for example:
`sprite.setShaderTexture("uNoise", noiseTexture);`

or:

```
sprite.shaderTextures = {
	uNoise: noiseTexture,
};
```

Then melonJS could handle upload/cache/bind and sampler uniforms internally, instead of user code touching raw WebGL texture units.

**Describe alternatives you've considered**

The current workaround is manually creating WebGL textures, binding them to texture units, and setting sampler uniforms by hand. It works, but it feels outside the normal melonJS texture pipeline and can get messy with custom shader/editor workflows.

**Additional context**

This would make procedural textures like NoiseTexture2d easier to use in custom shaders, not only as Sprite.image or normalMap, but also as extra inputs for effects like UV distortion, dissolve masks, flow maps, etc.

obiot:
Two additive, backward-compatible helpers on the base `ShaderEffect`, both aimed at custom shader animation/inputs — neither changes any existing effect.

## `ShaderEffect.setTime(seconds)`
A convenience for the `uTime` shader-animation convention, promoted to the base class. Call it once per frame to write a shader's `uniform float uTime` (scroll a static texture's UVs, pulse, wave), driven by whatever clock you choose (real / paused / scaled / scrubbed).

- No-op when the shader doesn't declare `uTime`, so it's safe on any effect.
- **Manual by design** — the engine never calls it; animation stays opt-in (exactly like re-baking a `NoiseTexture2d` with `update(dt)`).
- The built-in `Hologram` / `Shine` effects now **inherit** it — their identical `setTime` overrides were removed.

## `ShaderEffect.setTexture(name, image[, repeat])` — closes #1532
Bind an **extra** `sampler2D` to a custom post-effect, beyond the sprite/target it processes (`uSampler`) — a noise map, mask, gradient, or flow/lookup table.

```js
const noise = new me.NoiseTexture2d({ width: 256, height: 256, seamless: true });
const water = new me.ShaderEffect(renderer, `
    uniform sampler2D uNoise;
    uniform float uTime;
    vec4 apply(vec4 color, vec2 uv) {
        vec2 flow = texture2D(uNoise, uv + uTime * 0.03).rg - 0.5;
        return texture2D(uSampler, uv + flow * 0.02);
    }`);
water.setTexture("uNoise", noise.getTexture(), "repeat");
waterSprite.shader = water;
// each frame:
water.setTime(me.timer.getTime() / 1000);
```

The engine uploads the texture once, caches it, and (re)binds it to a reserved texture unit each draw while pointing the sampler uniform at it — **no raw WebGL texture-unit juggling** (the workaround #1532 asks to remove).

**How it works.** The bind is hooked at the single point both post-effect paths pass through, right after `useShader()`:
- single-effect sprite shader → `WebGLRenderer.drawImage` (`customShader` path, no FBO)
- multi-effect / camera → `QuadBatcher.blitTexture` (FBO blit)

Reserved units are the **top** of the batcher's range, so they never collide with `uSampler` (unit 0 on a blit) or the low, rotating color-texture pool. Extra textures are released in `ShaderEffect.destroy`.

**API note vs. the issue:** #1532 proposed `sprite.setShaderTexture(...)`; this puts it on the `ShaderEffect` instead — the shader that declares the sampler owns its inputs, and it's unambiguous when a sprite has several effects. It's still reachable as `sprite.shader.setTexture(...)`.

Enables UV-distortion water, dissolve masks, flow maps, palette lookups, etc.

## Tests
- `shadereffect-settime.spec.js` — writes `uTime` when declared + chainable; no-op otherwise.
- `shadereffect-settexture.spec.js` — **end-to-end readPixels** proving the extra texture is actually sampled (red extra over a blue source → red output), plus reserved-unit / sampler-uniform / upload-once-caching / destroy-frees assertions.
- Full render-path regression sweep green (shader, batcher, tmx-shader, depth, camera, container, adversarial pipeline, sprite). `eslint src tests` + `tsc` clean.


obiot:
Added a live **Water Refraction** example (`/water-refraction`) that dogfoods `setTexture`: a tiled pool floor + sunk "melonJS" watermark refracted through a **static** seamless `NoiseTexture2d` bound as an extra `uNoise` sampler, scrolled on the GPU via `setTime` — no per-frame CPU noise re-bake, no `Light2d`. Pointer-driven swells, animated caustics, and a live ripple-strength slider (`setUniform`). It sits alongside the existing **Water Surface** example, which demonstrates the complementary `Light2d` normal-map lighting approach.

obiot:
Thanks for the review — all points addressed in 6abe38bf5 (full suite 4547 passed, lint + tsc clean):

**`_usesTime` false-positive → `setUniform` throws** (shadereffect.js:111, x2). Dropped the `fragmentBody.includes("uTime")` substring scan entirely. `setTime` now detects `uTime` from the **compiled program's active uniforms at call time** (`typeof this._shader.uniforms.uTime !== "undefined"`). This is immune to `uTimeScale`/comments/optimised-out `uTime`, and — because it's evaluated per-call rather than cached — it also stays correct across a context-loss recompile (and no longer reads a null uniforms map when an effect is constructed during a lost-context window).

**Context loss invalidates cached `entry.tex`** (shadereffect.js:123). `_onContextLost` now nulls each extra texture's handle **and** releases its unit reservation, so `_prepareTextures` re-uploads + re-reserves on restore instead of binding a stale handle.

**No capacity check in `_prepareTextures`** (shadereffect.js:246). Units are handed out top-down and guarded — never below unit 1 (so no collision with `uSampler`), with a one-time `console.warn` if an effect binds more extra textures than the batcher has units.

**Sprite `customShader` unit collision** (webgl_renderer.js:1192). Root-caused to the `TextureCache` allocator not knowing about the directly-bound extra units. Fixed at the source per your suggestion: extra units are now **reserved in `TextureCache`** (reference-counted `reservedUnits`), and `allocateTextureUnit` skips them — so a sprite's own texture can never be handed a reserved extra unit, in either the blit or the customShader path. Released on destroy / rebind / context loss.

**Tests asserting private `_usesTime`** (settime spec, x2). Removed the private-flag assertions (contract-focused now) and added a `uTimeScale` false-positive regression (asserts `setTime` is a silent no-op and does not throw). Also added a unit-reservation regression proving the allocator skips a reserved unit and frees it on destroy.

PR #1538, opened 2026-07-04
feat(Sprite3d): anchor support via settings.anchorPoint + named presets engine-wide

### LINKED ISSUE #123: Create API for adding custom mouse and touch events
Since MelonJS has limited mouse and touch support it would be nice to have an api for adding custom mouse and touch events such as via a plugin or one's own code.

This would make it so the user doesn't have to hack the input.js source file in MelonJS and the custom events could be shared.

### LINKED ISSUE #456: Viewport should support multiple bounding regions
See: #349 and https://groups.google.com/forum/#!msg/melonjs/T7nOCwaI8vI/VqdsDkp0EhIJ

Multiple bounding regions will allow customized map scrolling mechanics found in games like Megaman, and Legend of Zelda: A Link to the Past. Where a map's scrolling will be "locked" into some region until the followed entity leaves the viewport.

### LINKED ISSUE #1514: Sprite3d: bottom/origin anchor for ground placement (2.5D)
## Context
`Sprite3d` (added in #1513) renders a **centered** quad. For 2.5D characters (Paper Mario-style), you almost always want the sprite's **feet on the ground**, which currently means manually offsetting `pos.y` by half the height.

## Proposal
Add an anchor / origin option so the pivot can be placed at the bottom-center (feet), e.g. `originPoint: { x: 0.5, y: 1 }` or a simple `anchor: "bottom"` shorthand, applied as a constant offset baked into the local quad (so it composes with billboarding/flip and doesn't affect the cull bounds).

- Default stays centered (no behavior change).
- Should work across all billboard modes and with trimmed/rotated atlas frames.

## Acceptance
- `new Sprite3d(x, y, { ..., originPoint: { x: 0.5, y: 1 } })` plants the bottom edge at `pos`.
- Unit test: corner Y offsets shift by the origin; billboard projection still upright.

Refs: docs note in [2.5D Games](https://github.com/melonjs/melonJS/wiki/2.5D-Games#paper-thin-characters-with-sprite3d) ("Ground anchoring" caveat).

asadullahbro:
Adds a settings.originPoint option (preset string or {x,y}) so the quad's pivot can be placed anywhere relative to pos, e.g. bottom-center for ground-anchored 2.5D characters. Baked as a constant local offset applied after flips, so it composes with billboarding and trimmed/rotated atlas frames. Default stays centered — no behavior change for existing code.

Updated the billboard example to anchor the character at its feet instead of using a manual GY offset fudge factor.

Closes #1514

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactoring (no functional changes)

## Checklist

- [x] My code follows the existing code style (`pnpm lint` passes)
- [ ] I have tested my changes locally (`pnpm test` passes)
- [ ] I have added tests that cover my changes (if applicable)


obiot:
Hi,

Thanks for the PR, before merging I'm gonna make some further changes however that I think makes it a better fit with the legacy/existing API :

1. drop originPoint and override the existing anchorPoint instead, so Sprite3d anchors exactly like the 2D Sprite:
   - accept settings.anchorPoint (same key/convention/default as Sprite; your preset strings can stay as sugar)
   - hook the inherited ObservablePoint via setCallback() so runtime changes re-bake the vertices, like the live 2D anchor (but then force `applyAnchorTransform = false` on both camera paths, so that vertex bake is the single anchor mechanism and nothing leaks into the view matrix)

2. keep the preset strings you added, but extend them to regular sprite (2d) object.

asadullahbro:
@obiot thanks for the detailed feedback! Happy to implement this myself, just let me know if you'd rather I push these changes, or if you're planning to handle it on your end. Either way works for me.

obiot:
> @obiot thanks for the detailed feedback! Happy to implement this myself, just let me know if you'd rather I push these changes, or if you're planning to handle it on your end. Either way works for me.

no worries, I will do it, already on it :)

obiot:
Thanks again @asadullahbro — the vertex-bake mechanism you built was verified correct end-to-end and is kept as-is. I've pushed the rework we discussed on top of your commit (`d06eb59be`):

**API change:** `settings.originPoint` → `settings.anchorPoint` — Sprite3d now anchors through the same key, convention and centered default as the 2D `Sprite`, and the inherited `anchorPoint` ObservablePoint is live (`sprite3d.anchorPoint.set(...)` re-bakes the quad and re-derives the cull bounds at runtime). A new `_anchorBaked` opt-out in `Mesh.preDraw` keeps the anchor out of the renderer transform on both camera paths.

**Your presets went engine-wide:** the names now work everywhere `settings.anchorPoint` is accepted — `Sprite`, `Entity`, `Collectable`, `ImageLayer`, `Text`, `BitmapText` and subclasses — via a shared resolver. 2D classes stay fully backward compatible (invalid values keep the historical silent `(0,0)` outcome, now with a console warning); Sprite3d throws on invalid input.

**Also fixed along the way:**
- cull bounds: a bottom-anchored 130×225 character had a ~47% under-sized cull sphere (visible pop-in); the bounds now grow to exactly enclose the anchored quad in any billboard orientation, byte-identical for the centered default
- example: the camera target (`GY * 0.6`) had become dead code aiming at the floor after `GY = 0` — framing restored to ≈ the original
- Sprite now forwards the *resolved* anchor into the texture-cache descriptor, so a preset string (or an aliased Vector2d) never lands in the shared per-image atlas

**Tests:** new `tests/anchorpoint-presets.spec.js` (cross-renderable contract, 72 tests) + 9 Sprite3d geometry/runtime/cull tests — all written failing-first. Full suite 4698/0. Master is merged in, so the branch is current.

obiot:
Follow-up hardening pushed (`51203d263`, `6d7e0486e`) — the anchor surface is now adversarially covered:

- **142 anchor tests total** across `anchorpoint-presets.spec.js` (93) and `sprite3d.spec.js` (49): all 9 presets value-validated, combined flip + frame-change + runtime re-anchor torture, rotated+trimmed regions, out-of-range anchors vs the cull sphere (property invariant), spherical rotation-invariance, Tiled string-property chain, `Text` pool recycling, and a regression guard pinning that the legacy bounds under-enclosed a bottom-anchored quad.
- **Backward-compatibility proof**: the whole spec was run against master's engine source — all 31 pure back-compat pins pass there (including a pin of the pre-existing shared-atlas pivot inheritance); the only deliberate divergence is that a caller's `Vector2d` is no longer aliased into the shared texture cache.
- **Example sweep**: platformer, sprite, aseprite, text, isometric-rpg, ui, tiled-map-loader and texture-packer all verified running in-browser with zero console/page errors.

Full suite: 4727 passed / 0 failed.

obiot:
all done, thanks again for the PR !

PR #1592, opened 2026-08-18
Physics: per-shape collision filtering, triggers and enable flag (#1590)

### LINKED ISSUE #1590: Physics: per-shape collision filtering, alongside the existing per-body type/mask
## Requet

A body can carry **several shapes**. Filtering is currently per **body** only — `collisionType` / `collisionMask` on `Body` — so every shape a body owns collides with exactly the same set as every other shape it owns.

Their editor lets each shape carry its own type and mask, so one body can mix, say, a hurtbox that only interacts with attacks and a solid footprint that only interacts with terrain. The body-level fields stay for compatibility; the shape-level filter is additional.

## Where it belongs — not `shouldCollide`

`shouldCollide(a, b)` (`physics/builtin/detector.js:268`) takes **renderables**, not shapes, and is called once per candidate pair from the broadphase (`:360`) *before* any shape pair exists. A per-shape rule cannot be expressed there. The best it could do is "could ANY shape pair match", which still needs the real test later, so it would be a duplicate not a home.

The right place is the nested shape loop in `collides()` (`:293-317`), immediately before the `SAT_LOOKUP` dispatch — the same spot that already `continue`s for an unlisted shape-type pair. A rejected pair simply never reaches SAT:

```js
for (shapeA of bodyA.shapes) {
  for (shapeB of bodyB.shapes) {
    if (!shapePairAllowed(shapeA, shapeB, bodyA, bodyB)) continue;   // <- here
    const test = SAT_LOOKUP[shapeA.type + shapeB.type];
    ...
  }
}
```

## Compatibility

Make the shape fields **inherit from the body when unset** rather than defaulting to a value:

```js
const typeA = shapeA.collisionType ?? bodyA.collisionType;
const maskA = shapeA.collisionMask ?? bodyA.collisionMask;
```

With no shape setting either, every pair resolves exactly as today — the existing body-level check is unchanged and no existing game shifts behaviour. This is purely additive.

## Points to settle

**Does the body-level check stay as a pre-filter?** Keeping `shouldCollide`'s existing body test means a body pair rejected there never reaches the shape loop, so a shape could not *widen* what its body allows — only narrow it. That is probably the behaviour to want (the body stays the coarse gate, shapes refine it), but it should be a decision rather than an accident, and it needs documenting either way. Allowing widening means relaxing `shouldCollide` to a union of its shapes' types/masks, which costs a cached recompute whenever shapes change.

**Hot-loop cost.** The check runs per shape pair, inside the innermost loop of the physics step. Two `??` and two ANDs is cheap, but bodies with several shapes multiply it. Resolving each shape's effective type/mask once when it is added to the body — rather than per test — keeps the loop the same shape it is now.

**Where the fields live.** Shapes are plain geometry (`Rect`, `Polygon`, `Ellipse`, `Line`, `Box3d`) and carry no physics state today. Adding two collision fields to every geometry gives every shape instance physics concerns it may never use, including the many created for rendering and bounds. A side table on `Body`, keyed by shape, is uglier to author but keeps geometry clean — worth weighing before picking.

**`Bounds`-only bodies.** A body with no shapes is already gated out by `shouldCollide`; confirm nothing changes for those.

## Testing

Cheap to pin, and the important cases are the ones that must NOT change:

- a body with two shapes and no shape-level filter behaves byte-identically to today (the compatibility guarantee)
- one shape of a two-shape body collides while the other is filtered out, and the response carries the right shape
- a shape-level filter cannot widen past its body's mask (or can, whichever is decided)
- filtered pairs never reach SAT — assert the narrowphase is not called, rather than only that no collision was reported

## Note

The `SAT_LOOKUP` dispatch immediately after the proposed insertion point was recently hardened (an unlisted shape pair used to throw mid-step and take the world update with it, now it warns once and treats the pair as no-collision). Anything added here should keep that property: a filter that throws is worse than one that over-collides.

### LINKED ISSUE #1591: Physics: a multi-shape sensor body is still pushed out (isSensor ignored by the junction-resolve pass)
Found while planning #1590 (per-shape collision filtering).

## What happens

`body.isSensor` means "detects contacts, is not solid". It works on a body with one shape, and **silently stops working the moment a second shape is added**.

```js
sensor.bodyDef = { isSensor: true, shapes: [new Rect(0, 0, 32, 32)] };
// -> holds position. Correct.

sensor.bodyDef = { isSensor: true, shapes: [new Rect(0, 0, 16, 32), new Rect(16, 0, 16, 32)] };
// -> pushed 8px out of the overlap. Wrong.
```

Measured: a sensor at `x = 100` overlapping a solid body by 16px is moved to `x = 92` after one `world.update()`.

## Cause

The main push-out honours the flag via its `eitherSensor` gate. But a body with 2+ shapes then runs an **extra-pass loop** (`physics/builtin/detector.js`, "resolve remaining overlaps at segment junctions") which exists to clean up leftover overlaps at shape junctions — polylines, compound colliders.

That loop writes positions **directly**:

```js
objA.body.ancestor.pos.set(objA.body.ancestor.pos.x - overlap.x * ratioA, ...)
```

rather than going through `respondToCollision`, and it gated only on `isStatic`. It never consulted `eitherSensor`, so it undoes the main path's decision.

## Why it was not caught

Nothing covered the intersection. The sensor specs (`builtin-adapter-collision-contracts`, `trigger`) all use single-shape bodies; the multi-shape specs (`body`) do not use sensors. Each half was tested, the combination was not.

## Fix

Gate the loop on `!eitherSensor`, alongside the existing shape-count condition.

## Impact

Anyone using a compound sensor — a multi-part trigger volume, a polyline detection zone, a hitbox built from several rects — has been getting push-out they explicitly asked not to have. Since it presents as an object drifting out of a trigger, it is more likely to have been worked around than reported.

Regression test added as `tests/sensor-multishape.spec.js`, with a single-shape control so a future failure distinguishes "sensors broke" from "compound sensors broke".

obiot:
Closes #1590. Closes #1591.

> **Five commits.** The per-shape feature, plus **three pre-existing bugs** found while building and reviewing it — each proven against `master` first, fixed in its own commit, and listed under `### Fixed` on its own terms rather than folded into the feature entry.

## The problem, plainly

A game object collides through a *body*, and a body can be built from several shapes — a box around a character's feet so they stand on floors, another around their torso so attacks can hit them. But the collision settings belonged to the *whole body*, so every shape on it collided with exactly the same things. Wanting "the feet only touch terrain, the torso only reacts to attacks" meant splitting into two objects with two bodies: awkward to author, and more to simulate.

Requested by a downstream maintainer shipping a customized melonJS runtime, who had already implemented it by **overriding `Detector` internals** in their wrapper — for want of an extension point.

## What each shape can now carry

| property | default | effect |
|---|---|---|
| `collisionType` / `collisionMask` | inherit the body's | this shape's own category / what it collides with |
| `isActive` | `true` | `false` removes it from collision entirely — no test, no contact, no events — without removing it from the body |
| `isTrigger` | `false` | `true` still collides and still fires events, but skips the position correction (a per-shape `isSensor`) |

Every default is the current behaviour, so **a shape that sets nothing behaves exactly as before**. A shape can only *narrow* what its body allows, never widen it — the body-level gate still runs first.

```js
const feet = new me.Rect(0, 24, 32, 8);
feet.collisionMask = me.collision.types.WORLD_SHAPE;
const torso = new me.Rect(0, 0, 32, 24);
torso.collisionMask = me.collision.types.ENEMY_OBJECT;
torso.isTrigger = true;          // detect hits, never get pushed by them
body.addShape(feet);
body.addShape(torso);
```

## A pre-existing bug, found and fixed first

Planning this surfaced a bug in the same code path, so it was **proven red, fixed, and committed separately** before any feature work.

`body.isSensor` worked on a body with one shape and **silently stopped working once a second was added** — a sensor overlapping a solid body by 16px was displaced 8px. Bodies with 2+ shapes run an extra pass to clean up overlaps at shape junctions, and that pass writes positions directly while checking only `isStatic`, undoing the main push-out's decision.

It survived because nothing covered the intersection: the sensor specs all use single-shape bodies, the multi-shape specs do not use sensors.

That fix had to land first — a trigger shape would otherwise have been correctly skipped by the main push-out and then repositioned by the loop anyway, and the feature would have looked broken for reasons predating it.

## Three things this had to get right

- **`addShape` does not store the object you pass.** A `Rect` and a `Bounds` become a *pooled* `Polygon`, and that pool's reset restores only position and vertices. Left alone, a recycled instance arrives still carrying the previous owner's mask or `isTrigger` and quietly changes how an unrelated body collides. `addShape` now writes every field explicitly, including defaults, idempotently (since `removeShape` re-adds survivors).
- **`??` and `=== false`, never `||`.** `collisionType: 0` is a legitimate value meaning "collides with nothing", and the flags are meaningfully false.
- **Position correction happens in two places**, not one — hence the bug fix above.

## Adapters

- **planck** — maps all four onto Box2D's native *per-fixture* filters and sensor flag, which it already built one of per shape. No structural change; `isActive: false` skips fixture creation. Bumped to **1.2.0**, melonJS `>=20.0.0`.
- **matter** — cannot: its `collisionFilter` and `isSensor` belong to a body, not to its parts. Documented as unsupported rather than half-emulated. Bumped to **1.1.1**, melonJS `>=20.0.0`.

## Three pre-existing bugs, fixed here

Each was found by writing a test that failed on `master` before any fix.

1. **A multi-shape sensor body was still pushed out** (#1591). `body.isSensor` worked with one shape and silently stopped working once a second was added — a sensor overlapping by 16px was displaced 8px. The junction-resolve pass for compound bodies writes positions directly while checking only `isStatic`. Had to be fixed first: a trigger shape would otherwise be correctly skipped by the main push-out and then repositioned by that loop anyway.
2. **`Body.destroy()` threw for an ellipse collider.** `addShape` accepts an `Ellipse`, but the release dispatch had no branch for one, so it hit the legacy pool and threw. `boundsPool.release` runs first, so teardown aborted partway and left the body holding a recycled `Bounds` — surfacing later as an unrelated broadphase crash. The same trap the `Box3d` branch beside it already documents.
3. **`Body.destroy()` threw when called twice.** The shape loop was already a no-op on the second pass, but `Bounds` and the vectors are set to `undefined` rather than emptied, so re-releasing threw "Instance is already in pool" — aborting teardown partway again.

## Tests

**250 files / 6059 passing**, plus planck 176 and matter 177. A new suite covers backward compatibility (a two-shape body with no settings; body-level masks; `Rect` vs `Polygon` equivalence), the new behaviour, and the adversarial cases — pool contamination, `removeShape` idempotence, falsy-is-not-unset, narrow-only, runtime toggling, all-shapes-inactive.

Mutations verified to break the suite: `??` → `||`, dropping the `isActive` guard, dropping the `addShape` transfer, removing the filter, and reverting the sensor fix.

**One honest gap:** two mutations survive — reading the trigger flag from the wrong shape index, and the extra-pass trigger guard. Both are unobservable through the public API, because `collides()` returns on the *first* colliding shape pair and iteration order decides which, so the reported pair cannot be steered to make the difference visible. The code is correct as written; a test that passed under mutation was removed rather than kept as decoration, and the guard is commented as defence-in-depth.

https://claude.ai/code/session_01QVjYzf76AEU3wJk766JAQi


obiot:
## Not ready — an adversarial review found a design-level flaw

Converting to draft. The per-shape `isTrigger` semantic does not work as documented, and the PR's own headline example is one of the cases that fails.

### The blocker

`Detector.collides()` returns on the **first** colliding shape pair. Push-out is therefore decided once per *body pair*, from one shape pair — so a per-shape trigger cannot be expressed without resolving every overlapping pair. Measured, identical geometry:

| shapes on the body | displacement |
|---|---|
| `[solid, solid]` | 24.000 |
| `[solid, TRIGGER]` | **0.000** |

With the trigger present, the **solid** shape is not resolved at all — it tunnels. Which shape wins depends on array order, because the shape loop walks in reverse. So `isTrigger` is documented as "a per-shape `isSensor`" but behaves as a whole-body `isSensor` selected by array position.

That invalidates the example in the JSDoc and in this PR body (feet solid + torso trigger on one body).

### Also found, not yet addressed

- `fromJSON` / `setVertices` / `addVertices` reach `this.shapes` **without** going through the normalizer, so the pooled-shape leak is open on the PhysicsEditor-JSON path
- the pass-through path (`Polygon`, `Ellipse`, `Point`, `Line`, `Box3d`) never clears `collisionType` / `collisionMask` on a recycled instance — the code comment claims it does
- the pool-recycling test is **vacuous**: `removeShape` and `removeChild` do not release to the pool (only `destroy()` does), so it passes for a reason unrelated to the code under test
- `isActive: false` does not remove a shape from raycasts, and inactive shapes still inflate `bounds`, `pseudoInertia` and `hasDepth`
- planck: every body-wide setter (`setCollisionType`, `setSensor`, …) flattens per-shape filters irreversibly; `body.collisionType` now reports a *shape* value order-dependently; a body with all shapes inactive has zero fixtures and is unrecoverable
- the modern `onCollisionStart/Active/End` responses carry no shape indices, so handlers cannot tell which shape was hit — the missing half of a per-shape API
- no planck-adapter tests for this feature at all, despite its CHANGELOG advertising support

### What stands

The first commit — the multi-shape sensor push-out fix (#1591) — is independent, proven red-then-green, and unaffected by any of the above. It may be worth splitting out and landing on its own.

obiot:
## Review findings addressed — `5a7478465`

### The blocker: a trigger no longer suppresses a solid sibling

`collides()` now **prefers a solid pair**. A trigger contact is remembered and the scan continues, so a solid pair wins whenever one exists. Only when no solid pair overlaps is the remembered trigger pair re-tested to repopulate the response and flagged via the new `ResponseObject.isTriggerContact` — which both push-out sites read, instead of re-deriving state from shape indices.

Measured, same geometry as the failure report:

| shapes on the body | before | after |
|---|---|---|
| `[solid, solid]` | 24.000 | 24.000 |
| `[solid, TRIGGER]` | **0.000** (tunnelled) | resolves |
| `[TRIGGER, solid]` | resolves | resolves |
| `[TRIGGER, TRIGGER]` | 0.000 | 0.000 |

With no triggers present the behaviour is bit-for-bit unchanged: the first hit still wins.

### Also fixed

- **`fromJSON` / `addVertices`** reach `this.shapes` via `setVertices`, not `addShape` — they now normalize too
- **Pool contamination** — `destroy()` scrubs the per-shape fields before releasing to the pools. That is the last point where the engine still knows the values are stale, and the only available fix for a shape handed straight to `addShape` with no separate source to correct from
- **`isActive` and raycasts** — both the 2D and 3D raycast paths now skip inactive shapes, matching the documented "removes it from collision entirely"
- **The vacuous pool test** — rewritten around `destroy()`, since `removeShape`/`removeChild` never release to the pool. It passed for a reason unrelated to the code under test

### Documented rather than changed

- `isActive: false` deliberately does **not** shrink the body's `bounds` — toggling a shape must not resize the body in the broadphase. Conservative bounds are safe; shrinking them is the risky direction.
- Under planck, the body-wide setters (`setCollisionType`, `setCollisionMask`, `setSensor`) write every fixture and so overwrite per-shape values, and `body.collisionType` reads the head fixture. Both are now called out in the adapter's CHANGELOG.

### Tests

**249 files / 6045 passing.** New cases: order-independence of a trigger sibling (both arrangements), the `isTriggerContact` flag set and cleared, raycast × `isActive`, static bodies, and `fromJSON` normalization.

Mutation-verified — each of these breaks the suite: reinstating the early return on a trigger pair, removing the pool scrub, and removing the raycast `isActive` check.

### Still honest about limits

Two paths remain unprovable through the public API and are commented as defence-in-depth rather than pinned: the `setVertices` normalization (the `destroy()` scrub keeps the pool clean, so it cannot be shown to change an outcome) and the extra-pass trigger re-check. Where a test could not be made to bite, I removed it rather than keep it as decoration.

Remaining known gaps, unaddressed: no planck-adapter tests for this feature, and the modern `onCollisionStart/Active/End` responses still carry no shape indices.

PR #1605, opened 2026-08-26
fix(input): account for world offset in pointer events

Vareniel:
## Description

Fix pointer event hit detection when `flex-height` or `flex-width` scaling introduces an offset on the root game world.

Previously, pointer coordinates and non-floating renderable bounds could use different coordinate spaces. This caused registered pointer events, such as `pointerdown`, to become offset from their visual renderables.

This change:

- Accounts for the root world offset when querying non-floating renderables.
- Preserves level-local coordinates for floating renderables.
- Uses the appropriate coordinate space for bounds checks and local pointer coordinates.
- Adds regression tests for vertical `flex-height` and horizontal `flex-width` world offsets.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactoring (no functional changes)

## Checklist

- [ ] My code follows the existing code style (`pnpm lint` passes)
- [x] I have tested my changes locally (`pnpm test` passes)
- [x] I have added tests that cover my changes (if applicable)

## Related issues

No related issue.


obiot:
Reviewed this, and the bug is real. Thanks for tracking it down — this one is nastier than the title suggests.

## Confirmed

`Camera2d.localToWorld` subtracts `world.pos`, so a pointer's `gameWorldX/Y` come out **level-local**, while a non-floating renderable's bounds are **absolute** and already include that offset. At the origin the two spaces coincide, which is why nobody hit this sooner. Move the world and the two disagree by exactly `world.pos`.

I checked the failure mode rather than assuming it, by reverting `pointerevent.ts` to master and rerunning: the handler is **never called at all**. Not a drift, not an off-by-a-few — total loss of input for every non-floating region. That is worth stating plainly in the changelog, which I have added.

## What I pushed to your branch

`maintainerCanModify` was on, so I have added coverage rather than asking you to write it. 15 new cases in `pointer-world-offset.spec.js`; **12 fail against master and pass with your fix**. The other 3 are the regression guards and pass either way, which is the point of them.

Beyond the two cases you had:

- negative, fractional and both-axis offsets
- a pointer **outside** a shifted region still misses. This was my main worry: the fix runs a second broadphase query and merges, so the hazard is a widened net that starts reporting hits nowhere near the pointer. Asserted with a positive control alongside, so it cannot pass by simply never hitting anything
- `gameWorldX/Y` stay level-local (the coordinates a game reasons in, not screen ones)
- **a child of a floating container.** `isFloating` inherits from the ancestor and your fix branches on it, so a child that never set `floating` itself must still take the floating path or a HUD built out of parts breaks
- a nested container with its own position, a scrolled camera, `pointermove`, and an offset changed after registration
- z-order between two overlapping regions, since the merged candidate list gets re-sorted and a dropped or duplicated entry shows up as the wrong region winning

## On regressions in other modes

No behaviour change: scale-method handling is identical before and after, and `world.pos` stays at the origin under all seven methods, at init and across a resize. `pointer-scale-modes.spec.js` pins that.

Which leads to the one thing I would flag about the framing: **nothing in the engine ever sets `world.pos`** — a game does, to centre a level. So this is not really specific to `flex-height`/`flex-width`; those just made it easy to notice. Any non-zero world offset reaches it. The tests therefore drive `world.pos` directly instead of going through a scale mode.

Deliberately not asserted: a click landing on a region under each scale method. Those call `renderer.resize()` against the parent element, which headlessly has no meaningful size (the canvas comes out 800x1731), so such a test measures the harness and not the engine. I verified that behaviour separately instead.

## Two things left

1. **`pnpm lint` did not pass** — biome wanted to reformat `input.spec.js`. Picked up in my commit, but worth running locally before the next push.
2. **Input-rate cost.** When the offset is non-zero the handler does a second broadphase query, a merge and a re-sort on **every** pointer event, including `pointermove`. Correct, and not something to hold this PR for, but a scene with a large candidate set will feel it. A follow-up could query once in the right space rather than querying twice and reconciling.

Also merged current master in, since 20.2.0 landed after you opened this. Full suite on the merged result: **6280 passed, 9 skipped**, types and lint clean.

PR #1608, opened 2026-08-28
perf(webgpu): memoize the per-quad segment slot lookup

obiot:
WebGPU submitted quads about **64% slower than WebGL 2** — 10.15 ms against 6.19 ms for 20 000 quads. The entire gap was one string built per quad.

## Where it was

`segmentSlotFor` runs once per quad to resolve which multi-texture slot the quad batches against:

```js
const slotKey = `${this.resourceId(record.view)}|${filter}|${wrap}`;
if (this.slotTable.peek(slotKey) === undefined) { ... }
```

Consecutive quads in a batch almost always share a texture — a sprite sheet, a font atlas, an emitter's particle image — so the answer was identical for long runs while the cost was not. At 20 000 quads that is 20 000 string allocations and 20 000 string-keyed map lookups **every frame**.

## It was not the GPU API

Worth recording, because the obvious suspects were all innocent. Instrumenting every WebGPU call in the draw path at 20 000 quads:

| call | per frame | ms |
| --- | ---: | ---: |
| `writeBuffer` | 21 | 0.05 (2.1 MB) |
| `createCommandEncoder` | 1 | 0.01 |
| `submit` | 1 | 0.01 |
| `createBindGroup` | **0** | 0 |
| `createBuffer` | **0** | 0 |

**0.07 ms of a 9.8 ms draw.** No bind-group churn, no per-frame allocation, and it uploads *less* than WebGL (2.1 MB vs 2.6 MB). The cost was plain JavaScript in the per-quad path.

## The fix

Remember the last resolved texture and short-circuit. Invalidated when the slot is evicted (the table hands that index to a different texture), and bypassed when the caller forces a `reupload`, since that means the resident record must be re-resolved.

| | before | after |
| --- | ---: | ---: |
| 20 000 quads | 10.15 ms | **5.71 ms** |
| 5 000 quads | 2.33 ms | **1.40 ms** |

WebGPU goes from noticeably slower than WebGL 2 to marginally faster (5.71 against 6.14). **This is the shared quad batcher**, so it lifts sprites, text and tilemaps too — particles are just what surfaced it.

Measured on Apple silicon under headless Chromium with ANGLE/Metal; absolute numbers will differ elsewhere, but the removed work is unconditional.

## Tests

6 added to `tests/webgpu_quad_batcher.spec.js`: repeated texture resolves to the same slot, distinct textures do not collapse onto one, `reupload` bypasses the memo, the memo does not survive a segment reset, an evicted slot is forgotten, and `addQuad` still routes quads to the right slot.

Two mutations, each breaking a test: making the memo ignore `reupload`, and dropping the eviction invalidation.

A third mutation found dead code rather than a gap — an explicit clear in `resetSegment` that no test could distinguish, because `slotTable.reset()` already evicts every live slot and the eviction callback handles it. Removed rather than left as untestable noise.

Full suite 6335 passing / 262 files, root lint 0 errors.


obiot:
**Update — the first version of this had two regressions, both caught by writing adversarial tests after the fact. Fixed in 088541695.**

The slot key is `(view, filter, wrap)`. My memo keyed on **texture identity alone**:

- The same texture drawn `nearest` then `linear` — or `no-repeat` then `repeat` — got the earlier slot back and batched against a sampler with the wrong configuration. Silent; visible only as a wrong-looking sprite.
- Worse, the memo short-circuited `getResidentRecord`, which revalidates the record against `texture.getTexture()` and re-uploads when the source changed underneath. A video frame, an animated canvas or a swapped atlas would have kept serving the stale upload.

The residency check now runs on every quad as before, and only the genuinely expensive part is memoized — building the template-literal key and hashing it into the slot table — keyed on the resolved view rather than the texture object, so a re-upload producing a new view misses.

**The correction costs almost nothing:**

| 20 000 quads | draw |
| --- | ---: |
| WebGPU, before any memo | 10.15 ms |
| WebGPU, unsafe memo | 5.71 ms |
| **WebGPU, corrected** | **5.93 ms** |
| WebGL 2, same run | 6.49 ms |

**Tests grew from 6 to 11**, and are now about the ways a memo can hand back a slot that is no longer correct rather than about the memo working when nothing changes: filter change on the same texture, wrap change on the same texture, interleaved textures each keeping their own slot, residency revalidated on every quad (counted, not assumed), `reupload` reaching the store, eviction driven through the real table by overflowing it, and interleaved quads sharing slots correctly through `addQuad`.

The two original regressions are directly covered — reverting either fix fails a test.

Issue #1404, opened 2026-04-19
GPU-instanced particle rendering


obiot:
## Summary

Render particles using GPU instancing instead of individual \`addQuad()\` calls. All particle state (position, velocity, scale, rotation, alpha, tint) is uploaded to a GPU buffer and rendered in a single draw call. Particle physics (gravity, fade, spin) is computed in the vertex shader — zero per-particle CPU cost per frame.

## Current State

- \`ParticleEmitter\` creates individual \`Particle\` renderables
- Each particle is drawn via \`addQuad()\` in the quad batcher — 4 vertices pushed per particle per frame
- CPU handles all per-particle updates: position, velocity, gravity, rotation, alpha fade, scaling
- Performance caps around 500-1000 visible particles before frame drops
- Each particle's \`update()\` and \`draw()\` is called individually

### Instance buffer

A single vertex buffer containing per-instance data for all active particles:
- \`vec2 position\` — world position
- \`float rotation\` — current angle
- \`float scale\` — current size
- \`float alpha\` — current opacity
- \`uint32 tint\` — color tint (packed ARGB)
- \`float life\` — remaining lifetime (0-1 normalized)

Updated once per frame via \`bufferSubData()\` — one upload for all particles.

### Vertex shader

Computes the final quad corners for each instance:
- Applies position, rotation, scale from the instance buffer
- Optionally computes physics on GPU (velocity, gravity, fade) if spawn parameters are uploaded instead of per-frame state

### Single draw call

\`gl.drawArraysInstanced(gl.TRIANGLE_STRIP, 0, 4, particleCount)\` — one call renders all particles regardless of count.

### Integration

- New \`GPUParticleEmitter\` class (or opt-in mode on existing \`ParticleEmitter\`)
- Shares the same configuration API: \`minLife\`, \`maxLife\`, \`speed\`, \`gravity\`, \`wind\`, etc.
- Falls back to existing CPU particle system for Canvas renderer
- Requires WebGL2 for \`drawArraysInstanced\` (or use ANGLE_instanced_arrays extension for WebGL1)

## Performance expectations

- **CPU**: near-zero per-frame cost — no per-particle update loop, no per-particle vertex pushing
- **GPU**: single draw call, single texture bind, hardware instancing handles the rest
- **Capacity**: 50,000-100,000+ particles vs current ~500-1000 cap
- **Memory**: one instance buffer (~24 bytes per particle) — 100K particles = ~2.4MB

## API Sketch

```js
[...]

PR #1611, opened 2026-08-28
test(bench): correct what the mesh benchmark claims gl.finish() does

obiot:
The mesh benchmark's comment asserts something measurably false:

```js
// gl.finish forces the driver to flush all queued GL work — without
// it, drawElements latency hides under the rAF cadence and the
// numbers look better than they really are.
renderer.gl.finish();
```

It does not, in the environment this benchmark runs in.

## Evidence

A/B on identical workloads, alpha-blended fill, median of 15 frames:

| `gl.finish()` | quads | megapixels | median |
| --- | ---: | ---: | ---: |
| no | 2 000 × 64² | 8.2 | 0.4 ms |
| **yes** | 2 000 × 64² | 8.2 | **0.4 ms** |
| no | 2 000 × 256² | 131 | 0.3 ms |
| **yes** | 2 000 × 256² | 131 | **0.3 ms** |
| no | 4 000 × 512² | 1 049 | 0.7 ms |
| **yes** | 4 000 × 512² | 1 049 | **0.7 ms** |

Over a **gigapixel of overdraw in a single frame at 0.7 ms** — about 1.4 TPix/s — with and without the call identical. Headless Chromium does not wait. A separate fill sweep across a 600× range of shaded pixels likewise stayed flat at 2.9 ms.

## What changes

Only the comment. The call stays (harmless, and it may synchronise in other environments), but it now says what the numbers actually are: **CPU-side submission cost**.

That remains the right measure for what this benchmark compares — the retained-mesh path wins by not re-uploading vertices and not re-transforming them on the CPU, and both are counted here. It just must not be read as end-to-end frame time, and adding the call back does not make it one.

## Scope

Every other `gl.finish()` in the suite sits immediately before `gl.readPixels()`, which synchronises on its own — redundant but correct, so they are untouched.

This came out of trying to establish whether the GPU is the bottleneck for particle rendering ([#1404](https://github.com/melonjs/melonJS/issues/1404)). It is not answerable with this instrument; doing it properly needs `EXT_disjoint_timer_query_webgl2` and real hardware rather than a headless proxy.

### DIFF
--- a/packages/melonjs/tests/drawmesh_bench.spec.js
+++ b/packages/melonjs/tests/drawmesh_bench.spec.js
@@ -190,9 +190,18 @@ describe("drawMesh benchmark (baseline for #1468)", () => {
 				draw();
 			}
 		}
-		// gl.finish forces the driver to flush all queued GL work — without
-		// it, drawElements latency hides under the rAF cadence and the
-		// numbers look better than they really are.
+		// NOTE: `gl.finish()` does NOT synchronise in this environment, so
+		// what follows is CPU-side submission cost, not end-to-end frame time.
+		// Verified by A/B: 4000 alpha-blended 512x512 quads — over a gigapixel
+		// of overdraw — measure 0.7ms with the call and 0.7ms without, which
+		// is not a rate any GPU achieves. Headless Chromium simply does not
+		// wait.
+		//
+		// That is the right measure for what this benchmark compares: the
+		// retained-mesh path wins by not re-uploading vertices and not
+		// re-transforming them on the CPU, and both are counted here. It just
+		// must not be read as a frame-time figure, and adding this call back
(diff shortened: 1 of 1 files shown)


obiot:
Superseded by #1612.

This PR corrected the comment to say `gl.finish()` does not synchronise, which was true but stopped one step short. The work it fails to wait for does not just go unmeasured — it escapes the test and drains in Chromium's shared GPU process, starving every page for minutes. That is the root cause of the CI failures, and #1612 fixes it properly: a 1x1 `readPixels` as a real barrier, so the cost lands inside the measurement, plus skipping the benchmark on CI where a software rasterizer measures the runner rather than the engine.

The comment this PR would have added is included there, with the mechanism rather than just the observation.

Issue #1515, opened 2026-06-23
Sprite3d / 2.5D: built-in blob shadow helper


obiot:
## Context

melonJS has no 3D shadow system. For 2.5D scenes (paper-thin billboard characters), characters read much better with a simple **blob shadow** on the ground. Today that's hand-rolled (a flat dark quad / `Sprite3d` with `billboard: false`, or a tinted `Mesh`) — see the "Shadows" caveat in [2.5D Games](https://github.com/melonjs/melonJS/wiki/2.5D-Games#paper-thin-characters-with-sprite3d).

## Why a fake shadow is the right answer here, not a shortcut

Worth stating explicitly, since "blob shadow" sounds like a placeholder for real shadows:

A shadow map would be **actively wrong** for this art style. A paper-thin billboard casts a paper-thin shadow — a line — which reads as broken rather than realistic. What the player needs from a shadow here is *contact*: where the character is standing, and how far off the ground they are mid-jump. A blob conveys exactly that.

It is also the wrong cost. Real shadows mean a depth pass per light, a shadow map texture, light-space matrices, and PCF sampling in the lit shaders on both backends — to end up looking worse for this case. Out of scope, and pointing the wrong way.

## Proposal — the sprite draws its own shadow

The technique is the blob; the design decision is **ownership**.

The obvious shape — a helper that spawns a paired shadow renderable and syncs it to its target each frame — is the part that would be a hack: two scene-graph nodes for one conceptual object, a depth-sort relationship to get right by hand, teardown that can leak, and a shadow that desyncs the moment someone moves the sprite without going through the helper.

Instead, **`Sprite3d` emits a second, ground-projected quad as part of its own draw**. It already emits one quad; a second costs no new render pass, no new node, and nothing to keep in sync.

```js
const hero = new Sprite3d(x, y, {
    image: "hero",
    castGroundShadow: true,      // or a settings object for size / opacity / fade
});
```

This removes two of the fiddly problems outright:

- **No z-fighting decision** — the sprite places its own shadow relative to itself, rather than a separate object guessing at the floor plane.
- **No sort-order problem** — both quads come out of one draw, so the shadow cannot land in front of the character.

Open sub-decisions:

- Size/opacity falling off with height above the ground (a jumping character's shadow shrinking and fading is most of the readability win).
- How the ground plane is established: a fixed Y, or a per-sprite `groundY`.
- A soft-edged radial texture (generated once, shared) versus an alpha-falloff in the shader.

## Acceptance

[...]

PR #1635, opened 2026-09-01
Mesh: a transparent pass, so a faded mesh actually fades

### LINKED ISSUE #1516: Mesh/Sprite3d: blended (alpha) pass for soft-transparent 3D sprites
## Context
The 3D mesh pass is **opaque** — `MeshBatcher.bind()` does `gl.disable(gl.BLEND)`. `Sprite3d` (#1513) handles transparency via an **alpha cutout** (`alphaCutoff`, default 0.5), which is correct and cheap for characters/foliage/signs (crisp silhouette, correct depth, no sorting).

But it can't do **smooth alpha blending** — soft particles, glows, fades, additive FX. Those need a blended, depth-sorted pass.

## Proposal
Add an opt-in blended path for meshes/`Sprite3d`:
- A blended draw mode (e.g. `blend: true` / a blend mode) that renders with `BLEND` enabled and `depthMask(false)` (depth-test on, depth-write off).
- Back-to-front ordering for the blended set (the existing depth sort can drive this), drawn after the opaque pass.
- Optional additive blend mode for glows.
- Per-sprite opacity/alpha fade (currently impossible since alpha is ignored).

## Considerations
- Transparent ordering is inherently order-dependent; document the limitation (intersecting blended sprites won't resolve per-pixel).
- Keep the opaque + cutout path the default (no perf regression for the common case).
- WebGPU port shape should stay clean (the batcher already owns mesh-mode GL state).

## Acceptance
- A semi-transparent `Sprite3d` (e.g. spherical particle) blends smoothly instead of hard-cutting.
- Per-sprite opacity fade works.
- Renders after opaque, sorted back-to-front; opaque/cutout path unchanged.

Refs: "soft alpha blending" out-of-scope note from #1513; [Working in 3D](https://github.com/melonjs/melonJS/wiki/Working-in-3D) transparency caveat.

obiot:
Closes #1516.

## The symptom, measured

`setOpacity(0.5)` did not make a mesh see-through. The mesh tier renders opaque — `MeshBatcher.bind()` disables `GL_BLEND` — and every mesh vertex shader premultiplies, so a faded mesh wrote `(rgb × a, a)` straight into the target.

A white mesh at half opacity over a **blue** background:

| | centre pixel |
|---|---|
| before | `[127, 127, 127]` — darkened toward black, background contributes nothing |
| after | `[127, 127, 255]` — blended |

That is the same defect as the far end of the range, which painted an opaque black silhouette until #1626 fixed it. The Canvas backend never had it: meshes there already fade under the 2D context, so the same scene faded on Canvas and darkened on GPU.

Nobody was relying on the darkening. The sledding example sets `rabbit.alpha = 0.35` for an invulnerability flash and has been getting a dark rabbit rather than a translucent one — visibly fixed by this branch.

## The change

Draws resolving to fractional alpha are queued and replayed after the opaque pass, back-to-front, blending, writing no depth but still depth-tested — so transparent objects composite with each other and stay correctly hidden behind opaque geometry.

```js
ghost.setOpacity(0.4);                                   // just fades
glow.transparent = true; glow.blendMode = "additive";    // texture alpha + additive
marker.transparent = false;                              // pinned opaque
```

Blending honours the renderable's **existing** `blendMode` — no new property.

**Ground shadows become the queue's first client** rather than the feature itself. `queueGroundShadow` / `flushGroundShadows` remain as a delegate and an alias, so `ground_shadow.spec.js` passes **52/52 with nothing but internals renames** — the regression gate for the whole refactor.

## Three defects found on the way, none of them the feature

1. **`beginBlendedDraw` double-multiplied alpha.** It passed `renderer.premultipliedAlpha` to the blend function, but that flag describes source *textures* — the mesh shaders premultiply unconditionally. A non-premultiplied context selected `SRC_ALPHA` and multiplied a second time. Invisible while decals were the only client, because their source colour is black and `0 × anything = 0`.
2. **`Renderer.reset()` leaked.** It zeroed the queue count without releasing entry references, keeping every queued renderable reachable until its pooled slot was reused. Harmless with two shared quads; not with arbitrary user meshes.
3. **`Sprite3d`'s `alphaCutoff: 0.5` would have defeated `transparent: true` outright** — discarding every soft texel before blending saw it, which is exactly the case the flag exists for. It now drops to `1/255` when transparency is explicit.

Also removes two fossils that would mislead the next reader here: a comment in `setBatcher` describing a drain #1630 deleted, and an orphaned JSDoc block documenting the method that went with it.

## Additive

An opaque scene pays two property reads and one compare against a packed tint **both backends already compute per draw**, plus three early-returning flushes per frame. No shader change. WebGPU pipeline keys are unchanged — the blend token and `|dw0` axis from #1515 already covered this, so `pipeline/cache.js` needed no edit at all.

## Tests

`tests/transparent_queue.spec.js` — 19 assertions: the measured symptom as an absolute pixel, the `true`/`false` escape hatches, order-independence of submission, near-on-top, not-painted-over-by-a-later-opaque-mesh, still-occluded-by-nearer-geometry, both #1630 drain traps, double-queueing with distinct matrices, destroy-before-drain, pool reuse and reference release, no re-queue during replay, mesh-mode state restored, and no blend-mode leak into the 2D cache.

Mutation-checked, all five bite: no sort (2 fail), reversed sort (1), screen-space guard removed (2), references not released (1), and the double-premultiply above (2).

`271 files / 6550 tests`, lint and types clean. Verified on screen on WebGL and on WebGPU/Metal, with the unfogged/untransparent examples unchanged.

## Out of scope, deliberately

Intersecting-transparency correctness (per-object sorting cannot order interpenetrating meshes — documented), the 2D-camera accumulated path, per-instance sorting inside an `InstancedMesh`, advanced destination-capture blend modes, and soft particles.


obiot:
### Follow-up commit — review fixes + per-target queues

A code review of the first commit found **five defects**, all fixed in `4616b2391`, plus two architectural changes that came out of it.

**The five:**

| | |
|---|---|
| A transparent `InstancedMesh` was routed into the pass and then replayed **opaque** on both backends | measured `[127,127,127]` — the set was deferred to end of frame and darkened anyway, strictly worse than never deferring it |
| `blendMode: "none"` threw mid-frame | a supported token with no blend state by contract; the replay dereferenced it |
| Entries replayed under the drain site's **live** view | a faded mesh inside a positioned container drew at the container's offset |
| The alpha cutout thresholded the **drawn** alpha | `Sprite3d` defaults the cutoff to 0.5, so a fade-out hard-cut to nothing at 49% opacity |
| The WebGPU replay read the mutable `premultipliedAlpha` | straight-alpha content earlier in the frame leaves it `false`, reapplying alpha a second time |

**One queue per render target.** Keyed on the shared render-target pool's active base — one number that means the same thing on both backends. A post effect binds its own target part-way through a scene, and a drain fired inside that bracket used to replay the *world's* queued geometry into the effect's offscreen buffer. Keyed this way the two cannot meet, and each pass drains its own on the way out.

**Overlays drain their own entries.** `Container.draw` now drains while the screen projection is still installed. Replayed after the bracket closed, entries went through the camera's perspective with vertices at view-space `z = 0` — the camera itself — and the divide deleted them: a faded mesh inside a `floating` child rendered **zero pixels anywhere in the frame**, while the same mesh at full opacity drew correctly. That was a regression introduced by the pass itself, so it is fixed rather than documented.

### API surface

Additive only, nothing removed or changed:

- **new public:** `Mesh#transparent` / `settings.transparent` (tri-state), forwarded by `Sprite3d`; `Renderer#flushTransparent()`
- **deprecated but working:** `flushGroundShadows()`, `queueGroundShadow()` — both delegate to the new names
- everything else (`queueTransparent`, `flushTransparentPass`, `transparentQueue`, `transparentTarget`, `resetScreenSpace`, …) is `@ignore`, so it never reaches the published `.d.ts`

Two **behaviour** changes existing games will see: a faded mesh now fades instead of darkening, and `alphaCutoff` thresholds the material's alpha so a fading cutout keeps its shape.

### Verification

- **6600 tests** (up from 6550), lint 0 errors, types clean
- every fix mutation-checked — reverted individually, confirmed the named test fails
- **all 50 examples run on both backends** in real Chrome (WebGPU/Metal, then again with `navigator.gpu` removed so the AUTO ladder falls to WebGL2 on the same driver): 100/100 clean, no blank canvases. The only backend divergence is `compressed-textures`, which is correct — WebGL has PVRTC, WebGPU does not

obiot:
Copilot's finding is correct and is now fixed in `24b7c9ded` — it landed on the same thing an independent review raised, which moved it from "residual, file an issue" to "fix it now".

The pass ordered its queue by squared distance from an eye recovered as `-Rᵀ·t`. That identity holds only for an orthonormal upper 3×3, and the matrix being read is **not the camera's view alone**: `Container.draw` folds every ancestor transform into `currentTransform`, so one scaled container makes the extraction wrong and silently mis-orders the pass — a transparent object composited over something actually in front of it.

Measured, with a view scaled `(3, 1, 1)` and translated `(0, 0, -1000)`:

| object | true view distance² | via `-Rᵀ·t` |
|---|---|---|
| at `(300, 0, 0)` | **1 810 000** | 1 090 000 |
| at `(0, 0, -200)` | 1 440 000 | 1 440 000 |

The first is genuinely farther; the extraction ranks them the other way round.

The fix pushes the position **through** the view rather than extracting the eye from it. That needs no assumption about the matrix, and agrees exactly with the old form whenever the view really is rigid, since a rotation preserves length — which is why every existing ordering test still passes unchanged.

Pinned by a test built on those measured numbers: restoring the old formula fails it with `expected 1090000 to be 1810000`.

One related spot deliberately left alone: `MeshBatcher.setPlacementUniforms` derives `uEyePosition` the same way for the specular half-vector. It has the same assumption, but it predates this PR, only affects specular highlight direction rather than draw order, and correcting it needs a real matrix inverse. Worth its own issue rather than widening this one.

PR #1639, opened 2026-09-03
Docs: stop publishing engine internals in the reference and in autocomplete

### LINKED ISSUE #1637: Docs: engine internals are published in the API reference and in consumers' autocomplete
## What's wrong

The published API reference lists a large number of members that are engine internals — and separately, the emitted `.d.ts` offers them in consumers' autocomplete. These are two different leaks with two different causes, and a member can hit either or both.

- **`@ignore`** hides a member from the docs (typedoc).
- **`@internal`** is what `scripts/strip-internal.ts` removes from the emitted `.d.ts`.

A member tagged only `@ignore` stays in the published types. A member with **no tag at all** appears in both.

The script's own header names the goal exactly: *"engine internals (pass lifecycle, texture retirement, …) would otherwise surface in consumers' autocomplete."*

## Scope

Counting only the unambiguous cases — underscore-prefixed fields, internal by this codebase's own convention, carrying no doc tag:

| class | untagged `_` fields |
| --- | --- |
| `Renderer` | 4 |
| `WebGLRenderer` | 22 |
| `WebGPURenderer` | 7 |

Examples: `_litPipelineWarned`, `_whitePixel`, `_activeTextureUnit`, `_effectPassDepth`, `_frameTexture`, `_currentGradient`, `_compressedTextureFormats`, and the eight `_advancedBlend*` fields (duplicated across both GPU backends).

Alongside them, untagged **internal methods** are documented as public API: `setBatcher`, `addBatcher`, `blitEffect`, `deleteMeshGeometry`, `setLightUniforms`, `clearRenderTarget`, `enableScissor` / `disableScissor`.

Non-underscore internal *fields* were not counted, and there are more — `lightUniformsScratch`, `maskDepthWarned`, `savedBlendMode`, `stubTextureView` are all in the reference today. `CanvasRenderer` was not audited.

## Why it matters

The reference is what someone reads to learn the engine, and the types are what their editor suggests while writing. Both currently present pass-lifecycle plumbing and one-shot warning flags as things to use. It also makes the genuinely public surface harder to find: the useful methods are buried among dozens that are not.

## Fix

Tag internals `@ignore` **and** `@internal` — the pattern already used in ~65 places. Mechanical, but it wants a member-by-member judgement call about what is genuinely public, so it should not be bundled into a feature PR.

Worth doing in one sweep across `Renderer`, `WebGLRenderer`, `WebGPURenderer` and `CanvasRenderer`, then extending to other classes: the renderers were audited because that is where this surfaced, not because they are unique.

## Already done

#1635 tagged the twelve methods and four fields it introduced, and verified they strip from the published types while `flushTransparent` and `flushGroundShadows` correctly remain. That is the pattern to follow; this issue is the rest of the surface.

obiot:
Closes #1637.

### The problem

Two leaks with the same symptom, and a member can hit either or both:

- **`@ignore`** hides a member from the generated docs — but leaves it in the emitted `.d.ts`.
- **A member with no tag at all** appears in *both*.

So the pass lifecycle, texture retirement, batcher plumbing and a long tail of one-shot warning flags were being offered to consumers as things to use. It also buries the genuinely useful API: the methods you want are lost among dozens you shouldn't touch.

`scripts/strip-internal.ts` already existed for exactly this — its own header names the case: *"engine internals (pass lifecycle, texture retirement, …) would otherwise surface in consumers' autocomplete."* It keys on `@internal`, and almost nothing carried that tag.

### The fix, in two halves

**Every `@ignore`d declaration now also carries `@internal`.** That is not a new judgement — it is the statement of intent the author already made, applied to the types as well as the docs. 1177 blocks.

**The stripper also drops class members whose name begins with `_`.** Tagging alone still left hundreds behind, because most internals were never tagged at all. The underscore prefix is this codebase's own convention for "not part of the API" and it is used consistently, so treating it as the declaration it already is covers the untagged ones *and* any added later — no tag needed on every one.

Deliberately limited to class members: a module-level `_name` is not emitted unless exported, and an exported one is a decision rather than an accident.

### Result

| | master | this branch |
| --- | ---: | ---: |
| underscore-prefixed members in the published types | 462 | **0** |
| declared members in the published types | 4747 | 3293 |

**1454 declarations removed, no public API moved.**

### Two deliberate exemptions

An `@ignore`d **type** that another emitted declaration imports cannot be stripped, or the published `.d.ts` is left with a dangling import — which breaks every consumer's type-check, not just the docs. `ResolvedApplicationSettings` and `Fog3dState` are both in that position: they keep `@ignore` and lose `@internal`, with a comment saying why so nobody re-adds it.

Both were found by type-checking the workspace plugins against the stripped output rather than by reading, which is the only reliable way to catch this class of breakage.

### Verification

- all five workspace plugins (`spine`, `debug`, `tiled-inflate`, `matter-adapter`, `planck-adapter`) type-check against the stripped declarations: **0 engine-type errors**
- the examples package likewise: **0**
- a spot-check that ~33 obviously-public members (`draw`, `preload`, `setOptions`, `play`, `addChild`, `drawMesh`, `setFog`, …) all survive
- full suite unchanged: **6633 tests**, lint and types clean

### Note on scope

This is mechanical and large (205 files) but behaviour-free: no runtime code changed except the build script. The diff is almost entirely one added `@internal` line per existing `@ignore` block.


obiot:
### Review fix

`d5187741a` — Copilot was right. `isUnderscoreMember` documented "class members only" but the implementation also matched `PropertySignature` / `MethodSignature`, i.e. **interface** members. An exported interface's `_field` is part of a contract someone may implement, so removing it changes that contract rather than hiding an implementation detail. `SpatialAudioState._pos`, `SpatialSoundState._pannerAttr` and nine siblings are real cases here. The check is now limited to class members, matching what the comment claims.

### Scope correction — this PR fixes one of the two leaks, not both

Worth stating plainly, because #1637's screenshots show the *other* one.

TypeDoc runs on **source** (`typedoc src/index.ts`), not on the emitted declarations — and it already honours `@ignore` by default. So adding `@internal` alongside every `@ignore` changes nothing in the generated reference. What it changes is the published `.d.ts`, which is a real and separate problem: those members were absent from the docs but present in consumers' autocomplete.

The members in the issue's screenshots — `_advancedBlendBusy`, `_effectPassDepth`, `lightUniformsScratch`, `savedBlendMode`, `setBatcher`, `blitEffect` — carry **no doc tag at all**, so they leak into *both*. This PR removes them from the types; they remain in the reference.

| | before | after |
| --- | ---: | ---: |
| underscore members in the published `.d.ts` | 462 | **0** |
| declared members in the published `.d.ts` | 4747 | 3293 |
| underscore members in the generated reference | 36 | 36 |

I attempted the reference half in the same branch and backed it out. The untagged members are mostly *lazily assigned* (`this._x = …` inside a method) rather than declared, so tagging them mechanically produces duplicated doc comments on fields that already have one — which TypeDoc reports as "multiple declarations with a comment" and resolves arbitrarily. My passes doubled the warning count (311 → 494) and in one case duplicated a description line inside an interface body. That wants a considered approach, not a regex.

Suggest merging this for the types fix and tracking the reference half separately — #1637 can stay open for it, or I can split it into its own issue with the above written up.

### Still green

6633 tests, lint and types clean, all five workspace plugins and the examples type-check against the stripped declarations with zero errors.

PR #1647, opened 2026-09-05
Level: an async option on load/reload/next/previous, and a microtask deferral

### LINKED ISSUE #1646: Level: add level.loadAsync(), and replace the 2011-era setTimeout deferral with a microtask
Two related changes to how a level load is scheduled and observed. Both are additive: no existing call form changes behaviour, and no shipped type changes.

## 1. The deferral is a 2011 artefact

`level.load()` stops the loop and defers the real work:

```js
if (wasRunning) {
	state.stop();
	setTimeout(() => {
		safeLoadLevel(levelId, options, true);
	});
}
```

That comment and its timer trace back to **2011-07-08 (v0.9.0)**, where it read *"pause the game loop to avoid some silly side effects"*. Promises are ES2015 and were not practical until roughly 2016; `queueMicrotask` arrived in 2018. So the timer was not chosen over a microtask — it was the only deferral primitive that existed. There is no macrotask semantic here worth preserving.

**The deferral itself is still needed.** `renderable/trigger.js:228` calls `level.load()` from a trigger handler, inside the game loop, potentially mid-iteration over the very container `safeLoadLevel` then calls `.reset()` on and destroys. `state.stop()` sets a flag but does not unwind the current frame. A microtask drains when the JS stack empties — the end of the rAF callback containing update *and* draw — so it unwinds the frame exactly as the timer does.

Swapping it also fixes a latent issue: `setTimeout` is clamped to ≥1000 ms in background tabs in every major browser, so a level load queued as the tab hides sits behind that clamp. A microtask has no clamp. The only thing the macrotask buys — yielding to the browser to paint once before the heavy work — is worth nothing here, because `state.stop()` has already halted the loop and `level.load()` does not switch to the loading state.

## 2. `level.loadAsync()`

Once the deferral is promise-shaped internally, the completion point exists and can simply be returned:

```js
function scheduleLoad(levelId, options, restart) {
	return Promise.resolve().then(() => safeLoadLevel(levelId, options, restart));
}
```

`load()` keeps `return true` and its `boolean` type; `loadAsync()` returns that same promise instead of discarding it.

**Why a sibling rather than changing `load()`'s return.** A Promise is truthy, so `if (me.level.load("x"))` would survive — but `build/level/level.d.ts` declares `function load(levelId: string, options?: {…}): boolean`, and this package ships its types. A TypeScript game doing `const ok: boolean = me.level.load("x")` stops compiling. `reload()`, `next()` and `previous()` delegate to `load()` and carry the same declaration. There is no single convention for this across asset-loading APIs — libraries designed Promise-first return the promise from `load()` directly, while those with an established non-Promise return add a sibling — and we are in the second situation. `loader.preload()` is already Promise-returning, so the engine ends up with both shapes, which is the price of not breaking typed consumers.

**The engine is already working around not having it.** `trigger.js:202` sequences fade-out → load → fade-in by monkey-patching the caller's own callback:

```js
const userOnLoaded = settings.onLoaded;
settings.onLoaded = function (levelId) {
	vp.addCameraEffect(new FadeEffect(vp, { direction: "out", … }));  // reveal
	if (typeof userOnLoaded === "function") userOnLoaded.call(this, levelId);
};
const onComplete = () => { level.load(gotolevel, settings); };
```

With a promise the wrap-and-restore disappears and the user's `onLoaded` is left alone:

```js
const onComplete = async () => {
	await level.loadAsync(gotolevel, settings);
	// re-read AFTER the load: `game.reset()` reassigns `app.viewport`
	// (application.ts:785), which is exactly why the wrapped callback
	// above reads it inside the callback rather than capturing it
	const vp = app.viewport;
	vp.addCameraEffect(new FadeEffect(vp, { direction: "out", … }));
};
```

That detail is load-bearing: the callback this replaces re-reads `app.viewport` at `trigger.js:174-175` with the comment *"re-read viewport after game.reset reassigns it"*. A refactor that captures the viewport before the load reintroduces a stale-viewport bug, so the promise form has to re-read it too — it removes the callback interception, not the need to read late.

**And it completes the game-side sequence**, which is already half promise-shaped since #1512:

```js
await loader.preload(assets);    // already awaitable
await level.loadAsync("map1");
```

## Scope notes

- The engine's preloader does not load levels and should not: `loader.js` and `loadingscreen.js` have no level references at all. Assets first, level after, by game code.
- Internal callers of `level.load()` are exactly two, both in `renderable/trigger.js` (lines 202 and 228), and both ignore the return value today.
- Failures should reject rather than throw asynchronously, so `await` can catch them. The existing synchronous `throw` for an unknown level id should stay synchronous — that is a programmer error, not a load failure.
- Related: #1644 wants a completion point to await a renderer warm-up at. That is a consumer of this, not a reason for it — the warm-up also works synchronously without any of this.

obiot:
Closes #1646.

## The timer

`level.load()` deferred its work with a timer so the current frame could unwind before the world is reset. **The deferral is still needed** — it is routinely called from a trigger handler mid-loop, and `safeLoadLevel` resets and destroys the very container the loop may be iterating, while `state.stop()` only sets a flag.

The timer itself is a 2011 artefact: that line and its comment trace to **v0.9.0**, four years before promises existed. Browsers clamp a timer to ≥1 s in a background tab, so a load queued as the tab hides sat behind that clamp. A microtask drains when the JS stack empties — the end of the rAF callback holding update *and* draw — so it unwinds the frame identically and is not clamped.

The no-loop branch stays **synchronous exactly as before**.

## The `async` option

`load`, `reload`, `next` and `previous` each keep one name and gain a flag:

```js
level.load("map1");                          // boolean, exactly as before
await level.load("map1", { async: true });   // Promise<boolean>
if (await level.next({ async: true })) { … }
```

`options.onLoaded` still fires either way, so the forms mix.

**No type break.** The signatures are preserved as JSDoc `@overload` pairs, *not* a `boolean | Promise<boolean>` union — a union fails every existing `const ok: boolean = level.load(id)` with TS2322, which I verified before choosing. The overload form compiles both that and the awaited call against the real emitted build:

```ts
const ok: boolean = level.load("map1");                       // still compiles
const a: boolean = await level.load("map1", { async: true });  // Promise<boolean>
```

Running out of levels still reports `false` rather than rejecting — reaching the end of a game is an ordinary outcome. An unknown level id throws **synchronously** in both forms: that is a typo, not a load failure, and it should not need `await` to surface.

The bounds check `next` and `previous` each spelled out is now a shared `levelIdAt(offset)` helper, so the two cannot drift.

## Trigger stops rewriting its caller's options

The fade/mask path sequenced hide → load → reveal by replacing `settings.onLoaded` with its own function and calling the user's from inside — mutating an option object the caller owns. Awaiting the load removes the interception.

The viewport is deliberately re-read **after** the load: `game.reset()` reassigns `app.viewport`, which is precisely why the callback this replaces read it late.

## The cost of putting the switch in the options

`await level.load(id)` without the flag is silent, because `await true` is valid JavaScript. It happens to be harmless today — the deferral is a single microtask queued before the await's continuation, so the load still runs first — but that is incidental ordering, not a contract. Documented on the `LevelLoadOptions` typedef and pinned by a test that records exactly this.

## Tests

No spec called `level.load()` at all before this, so both files are new — 24 tests across the legacy contract, the flag, the scheduling, the reload/next/previous paths, and the trigger.

Nine mutations of the changed behaviour fail as they should: the flag ignored, the microtask reverted to a timer, the unknown id rejecting instead of throwing, the loop not stopped, the synchronous branch throwing instead of rejecting, `next`/`previous` off-by-one, the `onLoaded` wrap reintroduced, and a stale viewport captured before the load.

A tenth (rewriting the bounds helper as `levelIdx[index] ?? null`) survives, and is an **equivalent mutant** rather than a gap: out-of-range array access is already `undefined`, so the explicit bounds check is belt-and-braces. The observable behaviour is covered by the two off-by-one mutations.

The viewport guard is a source check rather than a behavioural test — the reveal only runs when the hide tween completes, which needs a live game loop the suite does not have. It was vacuous on the first attempt (the explanatory comment in the inspected slice contained the string it asserted on), so comment lines are stripped before matching, and it was re-verified against the mutation.

## Also

`reload()` was documented as returning `object` — "the current level" — but returns whatever `load()` returns. The 2011 original returned nothing at all, so the declaration was never right. Corrected to `boolean`; `getCurrentLevel()` is the call that hands back the level object.

## Verification

276 files, 6716 tests, 0 failures. Lint 0 errors, build clean, emitted overloads and legacy compatibility both checked against the built types.


obiot:
Added the three missing twins — `reloadAsync()`, `nextAsync()`, `previousAsync()`. `loadAsync()` alone left a game able to await its first level but not a reload or a level transition.

**Each resolves with exactly what its synchronous counterpart returns**, so a port is mechanical:

```js
if (level.next())              →   if (await level.nextAsync())
```

That is why `loadAsync()` now resolves `true` rather than `void` — the rule is worth more than the marginally noisier type.

Running out of levels **resolves `false` without loading anything** rather than rejecting. `next()` returns `false` there, and reaching the end of a game is an ordinary outcome, not an error.

Emitted types, verified from `build/level/level.d.ts` — all four originals unchanged:

| | | |
|---|---|---|
| `load` | `boolean` | `loadAsync` → `Promise<boolean>` |
| `reload` | `object` | `reloadAsync` → `Promise<boolean>` |
| `next` | `boolean` | `nextAsync` → `Promise<boolean>` |
| `previous` | `boolean` | `previousAsync` → `Promise<boolean>` |

Six more tests, and five mutations of the new behaviour all fail as they should: `nextAsync` resolving true at the end, loading past the end, a `previousAsync` off-by-one at the start, `reloadAsync` not reloading, and `loadAsync` resolving undefined.

**Noticed but deliberately not fixed:** `reload()` declares `object` above, from a stale `@returns {object} the current level` — it actually returns whatever `load()` returns. Correcting it would change an emitted type, which is precisely what this PR is careful not to do, so it is left for a separate call.

Suite now 276 files, 6713 tests, 0 failures.

obiot:
Responses to the review, plus a design change since it was written.

**`queueMicrotask` → `Promise.resolve().then(() => { throw error })`** (two comments): declining. The two are not equivalent for this purpose, measured in the browser:

```
queueMicrotask(() => { throw e })          → ["uncaught-error"]
Promise.resolve().then(() => { throw e })  → ["unhandled-rejection"]
```

That rethrow exists specifically to preserve the *uncaught error* the timer-based version produced. The suggested form would turn it into an unhandled rejection, which is the behaviour change the code is there to avoid. On availability: `queueMicrotask` is baseline in every browser since 2018 and Node 11, comfortably older than the ES2022 target this package already requires.

**`loadAsync()` error-surface inconsistency**: valid, fixed. The deferred branch produced a rejection while the synchronous one let the exception escape the call, so the surface depended on `state.isRunning()` — and `.catch()` could never see the synchronous case, since the throw beat the handler being attached. The synchronous branch now returns a rejected promise, with a test asserting both branches reject identically. Good catch.

**`{@link ...}` in the changelog**: already gone — the entry had been rewritten before the review ran.

**PR description out of step with scope**: it was, and it is now further out of step, because the API changed after this review. The description has been rewritten.

**Design change:** the four `*Async` siblings are gone. Eight loading methods for four operations was too much surface, so the switch moved into the options object: `level.load(id, { async: true })`, and the same on `reload`/`next`/`previous`.

The signatures are preserved as JSDoc `@overload` pairs rather than a `boolean | Promise<boolean>` union — the union form fails every existing `const ok: boolean = level.load(id)` with TS2322, verified with tsc before choosing. The overload form compiles both the legacy call and the awaited one against the real emitted build.

obiot:
Second round.

**Options doc claiming the level is unloaded when you await without the flag** — correct, fixed. That is the same error I had already corrected in the skills and missed in this copy: with no loop the load is synchronous, and with one running the deferral is a single microtask queued ahead of the await's continuation, so it finishes first either way. The point still stands — without the flag there is no completion point to await — but the ordering is incidental, and the doc now says that rather than the opposite.

**Brittle source-substring test** — fair, fixed. Anchored on `load(gotolevel` instead of the full argument text, so reformatting or an added option cannot fail it. Re-verified that it still fails when the viewport is captured before the load, which is the only thing it exists for.

**`load()` delegating to `loadAsync()` losing the synchronous throw** — resolved, though by redesign rather than by patch. `loadAsync()` no longer exists: the switch moved into the options object, and the no-loop non-async path is now literally

```js
safeLoadLevel(levelId, options);
return true;
```

so a setup-time failure still throws synchronously and `try { level.load(id) } catch` still works. The promise form wraps the same call and rejects instead, which keeps the two consistent — that was the earlier review comment, and both are now satisfied at once.

**`queueMicrotask` fallback** — still declining, same reasoning as before. It is baseline in Chrome 71, Firefox 69, Safari 12.1 and Node 11, all from 2018–2019; ES2022 syntax this package already emits (class fields, private members) needs strictly newer engines than that, so there is no runtime that can parse the bundle but lacks the function. Adding a `setTimeout` fallback would also reintroduce the background-tab clamp on the error path this PR exists to remove. Worth restating that the choice is not stylistic: measured in-browser, `queueMicrotask(() => { throw e })` produces an **uncaught error** while `Promise.resolve().then(() => { throw e })` produces an **unhandled rejection**, and preserving the former is the point.

obiot:
Latest round.

**Stale `microtask` wording** (`level.js:199`, `:335`, `:374`, `:415`, and the spec header): correct, fixed. The deferral went back to a timer via `defer` and the docs did not follow. Good catch — that is exactly the kind of drift that outlives a revert.

**The reveal test only yielding to microtasks**: correct, and worse than it looked. The test *passed*, but `loaded` reached 1 during a microtask-only loop, which nothing in the test guarantees — it was measuring whatever the scheduler happened to do. It yields to a macrotask now. Both mutations it exists for still fail: a stale viewport captured before the load, and a reveal that does not wait for the load.

**Reveal ordering vs `state.restart()`**: the observation is right, the consequence does not follow, so no change.

The reveal did move from before `state.restart()` to after it. But it cannot produce a frame of unfaded level, because `state.restart()` schedules a `requestAnimationFrame` while the reveal is a promise continuation — a microtask in the same task the load ran in. Microtasks drain before the next animation frame, so the reveal is always in place before anything renders.

I wrote a test to pin that and then deleted it: asserting "no frame intervened" from outside the scheduler is racy, and the property is a language guarantee rather than engine behaviour. The existing test already pins the part that is ours — the reveal happens after the load, on the viewport that exists after it.

The other half of your point is real and intended: the caller's `onLoaded` now runs *before* the reveal rather than after. That is the interception being removed. The trigger no longer rewrites an options object it was handed, which is the defect this part of the PR exists to fix.

obiot:
**Note for anyone arriving from the squash commit:** its subject and body are stale. They were written against intermediate states that this PR later reverted, and the squash concatenated every commit message including the superseded ones.

What actually landed:

- The deferral is a **timer**, through the engine's own `utils.function.defer`. The microtask version was reverted: a macrotask cannot run inside another, so the load lands after the current frame structurally, rather than only while the update-and-draw path happens to stay synchronous.
- There is **no `loadAsync()`**. The switch is an `async` option on the existing `load` / `reload` / `next` / `previous`, preserved as JSDoc `@overload` pairs so `const ok: boolean = level.load(id)` still compiles.
- `queueMicrotask` is not used anywhere.
- The background-tab clamp is **not** fixed — that was the trade for going back to a timer, and the CHANGELOG entry claiming it was removed.

The code comments and the CHANGELOG in the merged tree are accurate; only the commit message is not. My miss for not retitling before the merge.

PR #1653, opened 2026-09-07
refactor: delegate font source resolution to the parser

### LINKED ISSUE #1648: Loader: fontface url() handling lives in the generic src-resolution path
`loader.load()` special-cases one asset type by name in the middle of the shared src-resolution path:

```js
// src/loader/loader.js:534
// strip url() wrapper for fontface assets so baseURL can be prepended to the raw path
if (asset.type === "fontface" && typeof src === "string") {
	const urlMatch = src.match(/^url\(\s*['"]?(.*?)['"]?\s*\)$/);
	if (urlMatch) {
		src = urlMatch[1];
	}
}
```

That is CSS knowledge sitting in the generic loader, and it is not the only piece: the `baseURL` guard immediately below also tests `src.startsWith("local(")` (`loader.js:546`), which is a CSS font descriptor form too. `data:` on the next line is genuinely type-agnostic; the other two are not.

### Why it cannot simply move to the parser today

The ordering forces it. `baseURL` is prepended at `loader.js:544-549`, and the parser only runs afterwards — so by the time `preloadFontFace` sees `data.src`, the prefix is already on. Without the strip, a descriptor written as `url('x.woff2')` would resolve to `"data/font/" + "url('x.woff2')"`, which the parser cannot repair after the fact.

So the result is that one value is unwrapped, prefixed, and then wrapped again in a different file:

```
loader.js    url('x.woff2')  ->  x.woff2  ->  data/font/x.woff2
fontface.js  data/font/x.woff2  ->  url('data/font/x.woff2')     (parsers/fontface.js:37-40)
```

with `url(` handled in both files and `local(` handled in one.

### Suggested shape

`parsers` is already a `Map` populated through `setParser(type, parserFn)` (`loader.js:180`, `:341`), so a parser can carry its own resolution rules and the loader can stop naming types:

```js
// parsers/fontface.js
preloadFontFace.resolveSrc = (src) => {
	const m = src.match(/^url\(\s*['"]?(.*?)['"]?\s*\)$/);
	return m ? m[1] : src;
};
preloadFontFace.skipBaseURL = (src) => src.startsWith("local(");
```

`load()` then does `src = parser.resolveSrc?.(src) ?? src` before prefixing, and consults `parser.skipBaseURL?.(src)` in the guard, keeping only the `data:` test inline. Both pieces of CSS knowledge end up in the file that understands CSS font descriptors, and the shared path becomes type-agnostic.

Naming is open — a single `resolveSrc(src, baseURL)` returning the final value would also work, at the cost of each parser repeating the prefix logic. The two-hook split above keeps `baseURL` resolution in one place.

### Notes

- Behaviour-preserving refactor: no asset should resolve differently afterwards.
- Existing coverage to keep green: `tests/fontface-src.spec.js`, `tests/loader-asset-src.spec.js`, `tests/asset-paths-with-spaces.spec.js` (a font path containing a space is the case the quoting exists for), and `tests/loader.spec.js`.
- Worth checking while in there whether any other parser would benefit from the same hook, or whether fontface stays the only user — if it is the only one, that is an argument for keeping the hook minimal rather than general.

ICOM725:
## Description

Move CSS font source handling out of the generic loader and into the fontface parser. The loader delegates source normalization and base-URL exclusion to optional parser hooks; the fontface parser owns `url()` unwrapping and `local()` handling.

Keep the existing matching rules and immutable asset descriptors. Tests register the font rules under a different parser name and cover bare paths, quoted/unquoted URLs, local fonts, data URIs and retries of the same frozen manifest entry.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Performance improvement
- [x] Refactoring (no functional changes)

## Checklist

- [x] My code follows the existing code style (`pnpm lint` passes)
- [ ] I have tested my changes locally (`pnpm test` passes)
- [x] I have added tests that cover my changes (if applicable)

Targeted verification: six loader/font spec files pass (98 tests), with Playwright using installed Chrome instead of Chromium Headless Shell through an external config override. Four of the seven new parser-alias cases failed before the change. `pnpm lint`, `pnpm build` and `git diff --check` pass.

The full suite was not run for this worktree. Full runs on the same upstream baseline with the independent Trigger change encountered browser-import/iframe failures and a stalled WebGPU spec; those runs are not reported as passing here.

## Related issues

Fixes #1648.


obiot:
Thanks for this — the direction is right and the delegation is the part worth keeping. I've pushed two changes on top of your commit (rebased onto master, which had moved ~20 commits): one to the shape of the hooks, one to their names.

### The shape

Your patch attaches `resolveSrc` / `skipBaseURL` as properties on `preloadFontFace`, which is what #1648 proposed. In practice that means `parsers.get()` returns a bare function for fourteen types and a decorated one for the fifteenth, so `load()` still has to probe for optional properties — it knows some types are special, just implicitly rather than by name. That is the thing moving CSS out of `load()` was meant to stop.

`setParser` now normalizes every registration into the same record:

```js
parsers.set(type, {
    parse: parserFn,
    normalizeSrc: parserFn.normalizeSrc ?? keepSrc,
    needsBaseURL: parserFn.needsBaseURL ?? alwaysRelative,
});
```

A type declares hooks exactly as you had it — on the parser function, in the parser's own file — and the fourteen that need neither write nothing. `load()` calls both unconditionally and never names a type. The contract is an `AssetParser` typedef next to `Asset`, matching how the loader documents itself elsewhere. `setParser(type, fn)` is unchanged for callers.

### The names

Both old names described what the *loader* should do rather than what the *value* is:

- `resolveSrc` → **`normalizeSrc`**. It resolved nothing — resolution is the base URL going on afterwards. What it does is turn a type-specific descriptor into a bare path
- `skipBaseURL` → **`needsBaseURL`**, inverted. "Skip a step" is an instruction to the caller, and read as a double negative at the call site. The positive form has a truthful default (yes, prefix it) and reads as a sentence: *normalize the src, then prefix it if it needs a base URL*

### A bug the rename exposed

`data:` and `skipBaseURL` were the same predicate — "this src needs no base URL" — one hardcoded, one delegated. Folding them naively means a type that overrides the hook **replaces** the data-URI rule, so a data-URI font gets a base URL prepended. `load()` now decides `data:` for every type *before* consulting the hook, so an overriding parser declares only its own exceptions and cannot forget that one. Your two data-URI cases would have caught it.

### Tests

Your seven cases are kept (renamed to the new hooks) and there are 23 more in `loader-parser-contract.spec.js`. The ones that earn their place are the ordering traps — a src that normalizes *into* a data URI, and one that normalizes into a form its own `needsBaseURL` then rejects — both of which resolve to a 404 if the exclusions are tested against the original src rather than the normalized one. Also pinned: a parser declaring no hooks is handed `url(...)` and `local(...)` verbatim, since those mean nothing to a type that has not said so.

Each piece verified by reverting it: folding `data:` into the default fails 5, running `normalizeSrc` after the prefix fails 10, applying it to an array src fails 1.

Your checklist noted the full suite was not run — it is green now: **287 files / 6951 tests**, `pnpm lint` clean, and the four specs #1648 listed as must-stay-green among them.

ICOM725:
Thanks for the follow-up! Having `setParser` fill in the defaults makes the loader simpler, and `needsBaseURL` reads much better at the call site. Keeping the data-URI check in the loader makes sense too, so individual parsers can't accidentally bypass it.

